PR #6, opened 2016-09-30
Cookiecutter format

jayoshih:


.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+# This file will be regenerated if you run travis_pypi_setup.py
+
+language: python
+python: 3.5
+
+env:
+  - TOXENV=py35
+  - TOXENV=py34
+  - TOXENV=py33
+  - TOXENV=py27
```
benjaoming: This will need some customization - I think it's possible to just start with 3.4 and 3.5.. the others are by all knowledge very rarely used since only very few Linux distros has ever shipped them.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,114 @@
...
+
+Get Started!
+------------
+
+Ready to contribute? Here's how to set up `ricecooker` for local development.
+
+1. Fork the `ricecooker` repo on GitHub.
+2. Clone your fork locally::
+
+    $ git clone git@github.com:your_name_here/ricecooker.git
```
benjaoming: The part that says `your_name_here` should be search/replaced with `learningequality`

docs/make.bat
```diff
@@ -0,0 +1,242 @@
+@ECHO OFF
```
benjaoming: This file is new to me :) I guess it's meant to support Windows devs, but it looks kind of too hard to maintain, and the Windows devs also have options to use contentional Make :)

setup.py
```diff
@@ -0,0 +1,59 @@
...
+    # TODO: put package requirements here
+]
+
+test_requirements = [
+    # TODO: put package test requirements here
+]
+
+setup(
+    name='ricecooker',
+    version='0.1.0',
```
benjaoming: I like the pattern where you import the package directly and just uses the version from there. So something like:

```
import ricecooker

setup(
    ...
    version=ricecooker.__version__
```

setup.py
```diff
@@ -0,0 +1,59 @@
...
+    install_requires=requirements,
+    license="MIT license",
+    zip_safe=False,
+    keywords='ricecooker',
+    classifiers=[
+        'Development Status :: 2 - Pre-Alpha',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: MIT License',
+        'Natural Language :: English',
+        "Programming Language :: Python :: 2",
```
benjaoming: TODO :)

tox.ini
```diff
@@ -0,0 +1,22 @@
+[tox]
+envlist = py26, py27, py33, py34, py35, flake8
```
benjaoming: Also TODO

docs/make.bat
aronasorman: The main dev is using Windows though :)

PR #30, opened 2016-12-06
Staging support

jayoshih:


ricecooker/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 __author__ = 'Learning Equality'
 __email__ = 'jordan@learningequality.org'
-__version__ = '0.3.5'
\ No newline at end of file
+__version__ = '0.3.5c12'
```
aronasorman: Is this the latest version? Maybe it's time to remove the beta-ness and upgrade to `0.3.6`!

ricecooker/classes/nodes.py
```diff
@@ -508,7 +509,7 @@ def validate(self):
         """
         try:
             assert self.kind == content_kinds.EXERCISE, "Assumption Failed: Node should be an exercise"
-            assert len(self.files) > 0 or len(self.questions) > 0, "Assumption Failed: Exercise should have at least one question or .perseus file"
+            # assert len(self.files) > 0 or len(self.questions) > 0, "Assumption Failed: Exercise should have at least one question or .perseus file"
```
aronasorman: Delete

ricecooker/classes/questions.py
```diff
@@ -28,6 +29,7 @@ def __init__(self, id, question, question_type, answers=None, hints=None, raw_da
         self.answers = answers if answers is not None else []
         self.hints = [] if hints is None else [hints] if isinstance(hints,str) else hints
         self.raw_data = raw_data
+        self.original_id=id
```
aronasorman: spaces around `=`

ricecooker/classes/questions.py
```diff
@@ -159,6 +169,87 @@ def validate(self):
         except AssertionError as ae:
             raise InvalidQuestionException("Invalid question: {0}".format(self.__dict__))
 
+    def process_question(self, downloader):
```
aronasorman: What's an example input for this function?

ricecooker/classes/questions.py
```diff
@@ -159,6 +169,87 @@ def validate(self):
...
+                    image_files += afiles
+
+        # Process raw data
+        self.raw_data = json.dumps(image_data, ensure_ascii=False)
+        self.raw_data, data_files = super(PerseusQuestion, self).set_images(self.raw_data, downloader)
+
+        # Return all files
+        return image_files + data_files
+
+    def process_image_field(self, data, downloader):
```
aronasorman: Why does `DownloadManager` need to be a separate parameter? How do you envision this to be used?

ricecooker/classes/questions.py
```diff
@@ -159,6 +169,87 @@ def validate(self):
...
+            Returns:string with checksums in place of image strings and
+                list of files that were downloaded from string
+        """
+        # Set up return values and regex
+        file_list = []
+        graphie_reg = re.compile(WEB_GRAPHIE_URL_REGEX, flags=re.IGNORECASE)
+        graphie_match = graphie_reg.match(text)
+        # If it is a web+graphie, download svg and json files,
+        # Otherwise, download like other files
+        if graphie_match is not None:
```
aronasorman: Can be simplified to `if graphie_match:`

ricecooker/commands.py
```diff
@@ -1,64 +1,174 @@
...
             path (str): path to file containing channel data
-            domain (str): domain to upload to
+            debug (bool): determine which domain to upload to
             verbose (bool): indicates whether to print process (optional)
         Returns: (str) link to access newly created channel
     """
+    # Get domain to upload to
+    domain = config.PRODUCTION_DOMAIN
+    if debug:
+      domain = config.DEBUG_DOMAIN
```
aronasorman: This only looks like two spaces.

aronasorman: Plus do we do anything else special with `--debug`, aside from setting `DEBUG_DOMAIN`? If not, let's remove it and just use `CONTENTWORKSHOP_URL`.

ricecooker/classes/questions.py
jamalex: I had the same question, but assumed it was part of your joint design. It seems a bit backward, perhaps -- might make more sense to have the tree get built up, and *then* passed to a stage that does the download process.

In terms of downloaders, we may want to consider using some similar machinery as in https://github.com/learningequality/kolibri/blob/develop/kolibri/content/utils/transfer.py (as it handles progress tracking, etc, well -- and also has separate processes for downloading and file copying). Perhaps those utils belong in le_utils...?

aronasorman: Doesn't have to happen now though. For now, can you answer the original question of how you plan to use the `DownloadManager` parameter?

jayoshih: Main reason for passing it around was to make sure the same downloader was used for all steps so that it would have one collective list of all of the files that had been downloaded. Probably would make more sense to change it to a global variable though

jayoshih: Will be moved to a separate module

PR #31, opened 2016-12-07
Img tag parser

jayoshih:
Parses img tags in questions so that they are properly formatted for exercise renderers


ricecooker/classes/nodes.py
```diff
@@ -204,11 +198,10 @@ def to_dict(self):
         """
         return {
             "title": self.title,
-            "description": self.description,
+            "description": self.description[:400],
```
aronasorman: Since we removed the limit on the cc server, why do we still have it here?

jayoshih: Changed in a different pr

PR #37, opened 2017-01-05
Add UnicodeDecodeError on dling skippable errors.

aronasorman:
This is the error that requests throws on Macs, if you give an invalid URL to requests, not InvalidURL.


jayoshih:
Looks good!

PR #44, opened 2017-01-19
Node ids

jayoshih:


ricecooker/classes/nodes.py
```diff
@@ -171,31 +181,39 @@ class ContentNode(Node):
...
+        self.extra_fields = extra_fields or {}
         super(ContentNode, self).__init__()
 
     def __str__(self):
         count = self.count()
         metadata = "{0} {1}".format(count, "descendant" if count == 1 else "descendants")
         return "{title} ({kind}): {metadata}".format(title=self.title, kind=self.__class__.__name__, metadata=metadata)
 
+    def get_domain_namespace(self):
+        return self.parent.get_domain_namespace()
```
jamalex: (Shared params like that can be handled in the base class and just passed through as part of `**kwargs`

ricecooker/classes/nodes.py
```diff
@@ -114,28 +114,24 @@ class Channel(Node):
             thumbnail (str): file path or url of channel's thumbnail (optional)
     """
     thumbnail_preset = format_presets.CHANNEL_THUMBNAIL
-    def __init__(self, channel_id, domain, title, description=None, thumbnail=None):
+    def __init__(self, channel_id, source_domain, title, description="", thumbnail=None):
         # Map parameters to model variables
-        self.source_domain = domain
+        self.source_domain = source_domain
         self.source_id = channel_id
```
jamalex: Having these match would be good too (in general, for any `__init__` arg that's directly applied as a class attribute)

ricecooker/sample_program.py
```diff
@@ -163,7 +163,7 @@ def construct_channel(**kwargs):
 
     channel = Channel(
         source_domain="learningequality.org",
-        channel_id="testing-ricecooker-channel",
+        source_id="testing-ricecooker-channel",
```
jamalex: Ah, yeah, this changes how people use it, too. We'll need to communicate the param name changes to the folks doing chefs.

jamalex:
Looks good! Could you update the version number, and then can someone (you, or @aronasorman?) push the latest to PyPI and let @ralphiee22 and the MagoGenie folks know what version to update their chefs to, and any changes they may need to make (e.g. the renamed args).

PR #45, opened 2017-01-20
File class refactor

jayoshih:
Dependent on https://github.com/learningequality/ricecooker/pull/42, https://github.com/learningequality/ricecooker/pull/43, and https://github.com/learningequality/ricecooker/pull/44


ricecooker/managers/tree.py
```diff
@@ -33,71 +33,33 @@ def process_tree(self, node, parent=None):
...
-            if isinstance(node, nodes.Exercise):
-                node.process_questions()
+        downloaded = node.process_files()
 
         # Process node's children
         for child_node in node.children:
-            self.process_tree(child_node, node)
+            downloaded += self.process_tree(child_node, node)
+
+        return list(filter(lambda x: x, set(downloaded))) # Remove any duplicate or null files
```
jamalex: Filters can be nice, but for some things, list comprehensions are more readable. This could also be:
`return [x for x in set(downloaded) if x]`

ricecooker/managers/tree.py
```diff
@@ -139,7 +100,10 @@ def reattempt_upload_fails(self):
             Args: None
             Returns: None
         """
-        self.upload_files(self.failed_uploads)
+        if len(self.failed_uploads) > 0:
+            config.LOGGER.info("\nReattempting to upload {0} file(s)...".format(len(self.failed_uploads)))
+            import pdb; pdb.set_trace()
```
jamalex: Remove pdb -- or is it intended to be here to help chefs?

ricecooker/managers/tree.py
```diff
@@ -161,10 +125,11 @@ def reattempt_failed(self, failed):
             config.LOGGER.info("\tReattempting {0}".format(str(node[1])))
             for f in node[1].files:
                 # Attempt to upload file
-                with  open(config.get_storage_path(f['filename']), 'rb') as file_obj:
-                    response = config.SESSION.post(config.file_upload_url(),  files={'file': file_obj})
-                    response.raise_for_status()
-                    self.uploaded_files += [f['filename']]
+                if f.filename:
```
jamalex: I think you already do validation and/or filtering on this above, so maybe OK to skip this check (or do it as an `assert` here)?

ricecooker/classes/files.py
```diff
@@ -0,0 +1,566 @@
...
+
+    def derive_thumbnail(self):
+        key = "EXTRACTED: {} (extracted thumbnail)".format(self.path)
+        if FILECACHE.get(key):
+            return FILECACHE.get(key).decode('utf-8')
+
+        config.LOGGER.info("\t--- Extracting thumbnail from {}".format(self.path))
+        with tempfile.NamedTemporaryFile(suffix=".{}".format(file_formats.PNG)) as tempf:
+            tempf.close()
+            extract_thumbnail_from_video(self.path, tempf.name, overwrite=True)
```
jayoshih: https://github.com/learningequality/pressurecooker/pull/5

ricecooker/classes/files.py
```diff
@@ -147,28 +72,38 @@ def write_and_get_hash(path, write_to_file, hash=None):
...
+    # Some files might have been closed, so only filepath will work
+    if isinstance(srcfile, str):
+        srcfile = open(srcfile, 'rb')
+
+    # Write file to local storage
+    with open(config.get_storage_path(filename), 'wb') as destf:
+        shutil.copyfileobj(srcfile, destf)
+
+    if delete_original:
+        os.remove(srcfile.name)
```
jamalex: If `delete_original` is set, it'd be much more efficient to move the file, rather than copy and delete.

ricecooker/classes/nodes.py
```diff
@@ -37,17 +37,21 @@ def guess_content_kind(files, questions=None):
 
 class Node(object):
     """ Node: model to represent all nodes in the tree """
-    def __init__(self, title, description=None, thumbnail=None):
+    def __init__(self, title, description=None, thumbnail=None, license=None, copyright_holder=None):
```
jamalex: I think we do need to make license a required field, to ensure they're being specified.

ricecooker/classes/nodes.py
```diff
@@ -37,17 +37,21 @@ def guess_content_kind(files, questions=None):
...
         self.files = []
         self.parent = None
         self.node_id = None
         self.content_id = None
+        self.title = title
+        self.description = description or ""
+        self.license = license
+        self.copyright_holder = copyright_holder
 
         if thumbnail and isinstance(thumbnail, str):
```
jamalex: What if it's already a `ThumbnailFile` instance? It should still add it.

ricecooker/classes/nodes.py
```diff
@@ -148,6 +150,9 @@ class ChannelNode(Node):
             title (str): name of channel
             description (str): description of the channel (optional)
             thumbnail (str): file path or url of channel's thumbnail (optional)
+            license (str): default license to use if nodes don't have license specified (optional)
```
jamalex: We now require license to be specified on each node, so channel-level license can be removed.

ricecooker/classes/nodes.py
```diff
@@ -203,25 +208,25 @@ class ContentNode(Node):
         Base model for different content node kinds (topic, video, exercise, etc.)
 
         Attributes:
-            id (str): content's original id
+            source_id (str): content's original id
             title (str): content's title
             description (str): description of content (optional)
             author (str): who created the content (optional)
             license (str): content's license based on le_utils.constants.licenses (optional)
```
jamalex: Would be good to use a class-based approach for licenses, as well, for consistency and to support validation. `copyright_holder` could then be an argument to the `License` object, and be required for `CCByNCSALicense`, for example, but not be required for `PublicDomainLicense`.

jamalex:
Looks good! We may want to do the mentioned "class-ification" of licenses before we ask chefs to upgrade, so we maintain better backwards compatibility, but your call, if it seems messy.

PR #50, opened 2017-01-26
Added version compatibility check

jayoshih:
(To address update on https://github.com/fle-internal/content-curation/pull/127)


ricecooker/commands.py
```diff
@@ -157,18 +163,35 @@ def prompt_token(domain):
...
+    result = json.loads(response._content.decode('utf-8'))
+
+    if  result['status'] == 0:
+        config.LOGGER.info(result['message'])
+    elif result['status'] == 1:
+        config.LOGGER.warning(result['message'])
+    elif result['status'] == 2:
+        config.LOGGER.error(result['message'])
+        if not prompt_yes_or_no("Continue anyways?"):
+            sys.exit()
```
jamalex: My thought was this prompt should go on status==1, and for status==2, it shouldn't allow them to continue at all (always exit). This allows us to ban misbehaving versions completely, which could be important.

jamalex: And both 0 and 1 should probably be "warning", not just "info".

jamalex: As described in: https://github.com/fle-internal/content-curation/pull/127#issuecomment-275214396

jayoshih: Think I might have used a slightly different numbering system:
0 = Success, Ricecooker is up to date
1 = Soft Warning, newer version of Ricecooker exists, but current version is fine too
2 = Hard Warning, user must confirm whether or not to continue
3 = Error, Ricecooker automatically exits

PR #52, opened 2017-01-27
Fixed validation logic to allow paths without an extension

jayoshih:
Ricecooker will use a default_ext if the path doesn't end with one. However, validation requires it to have an extension. This PR fixes that

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -194,8 +194,9 @@ class DownloadFile(File):
 
     def validate(self):
         assert self.path, "{} must have a path".format(self.__class__.__name__)
-        if os.path.splitext(self.path)[1][1:] != "":
-            assert os.path.splitext(self.path)[1][1:] in self.allowed_formats, "{} must have one of the following extensions: {}".format(self.__class__.__name__, self.allowed_formats)
+        _basename, ext = os.path.splitext(self.path)
+        if ext:
+            assert ext.lstrip('.') in self.allowed_formats, "{} must have one of the following extensions: {}".format(self.__class__.__name__, self.allowed_formats)
 
     def process_file(self):
         try:
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -116,7 +116,7 @@ def compress_video_file(filename, ffmpeg_settings):
 
         filename = "{}.{}".format(get_hash(tempf.name), file_formats.MP4)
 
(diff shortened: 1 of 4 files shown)


ricecooker/classes/files.py
```diff
@@ -207,28 +207,43 @@ class ThumbnailFile(ThumbnailPresetMixin, DownloadFile):
     default_ext = file_formats.PNG
 
     def validate(self):
-        assert os.path.splitext(self.path)[1][1:] in [file_formats.JPG, file_formats.JPEG, file_formats.PNG], "Thumbnails must be in jpg, jpeg, or png format"
+        super(ThumbnailFile, self).validate()
+        if os.path.splitext(self.path)[1][1:] != "":
```
aronasorman: I prefer this version:

```
basename, extension = os.path.splitext(...)
if extension:  # not empty
    ...
```

aronasorman: Other lines on this file have this same pattern, maybe you can change it?

ricecooker/classes/files.py
```diff
@@ -207,28 +207,43 @@ class ThumbnailFile(ThumbnailPresetMixin, DownloadFile):
...
+            assert self.path.endswith(file_formats.PDF), "Document files must be in pdf format"
 
 class HTMLZipFile(DownloadFile):
+    default_ext = file_formats.HTML5
+
     def get_preset(self):
         return self.preset or format_presets.HTML5_ZIP
 
     def validate(self):
```
aronasorman: These `validate` functions also all have the same general logic. Can you find a way to just have them use the base class's `validate` command?

ricecooker/sample_program.py
```diff
@@ -93,18 +103,18 @@ def guess_file_type(filepath, kind):
...
-                "path": "C:/users/jordan/Videos/testfolder/captions.vtt",
-                "language": languages.getlang('en').code,
-            }
-            ,
-            {
-                "path": "file:///C:/users/jordan/Videos/testfolder/captions.vtt",
-                "language": languages.getlang('es').code,
-            }
+            # {
+            #     "path": "C:/users/jordan/Videos/testfolder/captions.vtt",
```
aronasorman: Delete commented out code

jayoshih: Temporarily commented out since subtitles aren't allowed on the production server yet

PR #78, opened 2017-06-12
Sushi bar client: MMVP

antonio-cortes-perez:
Sushi bar client: MMVP features.

- Creates channels and runs.
- Reports stages.
- Sends logs.
- Reports statistics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ You will need an authorization token to create a channel on Kolibri Studio. In o
 
 ### Step 2: Initializing the Channel ###
 
-To run the Ricecooker, you must include a `construct_channel` method in your sushi chef file that returns a ChannelNode object. This function will be responsible for building the structure of your channel.
+To run the Ricecooker, you must include `create_channel` and `construct_channel` methods in your sushi chef file.
+`create_channel` should return only a ChannelNode object. `construct_channel` will be responsible for building the structure of your channel.
 
 Start by importing `ChannelNode` from `ricecooker.classes.nodes` and create a ChannelNode object. The ChannelNode class has the following fields:
 - __source_id__ (str): channel's unique id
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -35,7 +35,8 @@ You will need an authorization token to create a channel on Kolibri Studio. In o
 
 ### Step 2: Initializing the Channel ###
 
-To run the Ricecooker, you must include a `construct_channel` method in your sushi chef file that returns a ChannelNode object. This function will be responsible for building the structure of your channel.
+To run the Ricecooker, you must include `create_channel` and `construct_channel` methods in your sushi chef file.
```
ivanistheone: Why do we need both `create_channel` and `construct_channel`? Couldn't the `construct_channel` do both tasks as before?

ivanistheone:
I looked at the code briefly. Very clean. I left one question regarding the need for two channel functions.

I wonder if it's not better to keep the reporting/monitoring ricecooker in a separate branch for now, like `learningequality:ricecooker:monitoring_v1`?  or maybe just `learningequality:ricecooker:monitoring` and update the branch after MVP, etc.

README.md
antonio-cortes-perez: It could. But the idea of create_channel is just to create only the channel node (root node) as fast as possible, this is used to configure the sushi bar client as early as possible in upload_channel (this helps, among other things, to start reporting remote logs very early in the workflow). We discussed this with Jamie during last week meeting. construct_channel is the same as before, it can be a heavy weight call.

antonio-cortes-perez:
Thanks Ivan! Since, I need to start wrapping things up this week, I would prefer to merge to master. All the changes should be nop for existing sushi chefs, and it will only affect upgraded chefs (i.e. the ones with create_channel).

README.md
jayoshih: Is the main goal of this to get the channel id as soon as possible? It might be possible to just get the source_domain and source_id to calculate the channel_id (maybe have the user pass them in as parameters?)

jayoshih:
Looks good

PR #79, opened 2017-06-14
Remote logs via websockets

antonio-cortes-perez:


ricecooker/sushi_bar_client.py
```diff
@@ -168,33 +169,15 @@ class LoggingHandler(logging.Handler):
...
 
     def emit(self, record):
-        data = self.format(record)
         try:
-            response = requests.post(
-                config.sushi_bar_logs_url(self.run_id),
-                data=data,
-                auth=AUTH)
-            response.raise_for_status()
+            self.ws.send(json.dumps(record.__dict__))
```
ivanistheone: I think it makes more sense not to rely on the `ws:` path to pass the `run_id`. You could instead add it to the json payload as follows:

```
log_data = record.__dict__
log_data['run_id'] = self.run_id
json.dumps(log_data)
```

antonio-cortes-perez: Done

PR #80, opened 2017-06-14
Process Tree Optimization

miraziz:
Optimized tree file processing by using a global list instead of creating a large number of local lists.

### DIFF
--- a/ricecooker/commands.py
+++ b/ricecooker/commands.py
@@ -284,6 +284,7 @@ def upload_files(tree, file_diff):
     tree.reattempt_upload_fails()
     return file_diff
 
+
 def create_tree(tree):
     """ create_tree: Upload tree to Kolibri Studio
         Args:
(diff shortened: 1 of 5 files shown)


ricecooker/commands.py
```diff
@@ -292,10 +293,12 @@ def create_tree(tree):
     """
     # Create tree
     config.LOGGER.info("\nCreating tree on Kolibri Studio...")
+    # channel_id, channel_link = tree.upload_channel_structure()
```
ivanistheone: Is this meant to be commented out? i.e. we're not using `upload_channel_structure` now, but will be used in the future?

miraziz: Im currently committing changes in working segments, and will likely finish the remainder of this part today. I commented it out because that's the code that will need to be run instead of tree.upload_tree(), but the code on the CC side isn't done yet.

PR #90, opened 2017-06-26
ChefClasses and argparse command line args

ivanistheone:
Proposal for a new API for sushi chefs `<-->` ricecooker interactions.

Changelog:

  - rewrote commadn line argument parsingusing `argparse`
  - Defined two chef classes to manage `main` (entry point), and `run` (calls to uploadchannel)
  - Minimal changes to commands.py to support the new API
  - Use of `BaseChef(compatibility_mode=True)` to be backward compatible with old API
  - Basic tests to check argparse command line arg parsing matches docopt
  - Some minor cleanup `rieceooker/cli.py` and `Click` package were not used


tests/test_argparse.py
```diff
@@ -0,0 +1,178 @@
...
+
+@pytest.fixture
+def command_line_inputs(fake_chef_path):
+    test_input_templates = [
+      'ricecooker uploadchannel {} --token=letoken --resume --step=START_UPLOAD',
+      'ricecooker uploadchannel {} --token=letoken --reset somethin=else extrakey=extraval',
+      'ricecooker uploadchannel -uv {} --warn --compress --download-attempts=4 --token=besttokenever --resume --step=PUBLISH_CHANNEL --prompt --publish',
+      'ricecooker uploadchannel {} -v --compress --token=katoken lang=en',
+      'ricecooker uploadchannel {} -v --compress --token=katoken lang=en-us',
+      'ricecooker uploadchannel {} -v --compress --token=katoken lang=fr-ca',
```
ivanistheone: Anything else we should test?

docs/design.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+Summary
+-------
+A sushi chef script using the new API looks like this:
+
+    #!/usr/bin/env python
+    ...
+    ...
+    class MySushiChef(riececooker.chefs.SushiChef):
```
jayoshih: Typo?

tests/test_argparse.py
jayoshih: Maybe the update flag?

docs/design.md
```diff
@@ -0,0 +1,193 @@
...
+  - The `main` method of the class parses the command line arguments and calls
+    the `run` method (or the `deamon_mode` method.)
+
+        class BaseChef():
+            ...
+            def main(self):
+                args, options = self.parse_args_and_options()
+                self.run(args, options)
+
+  -  The chef's `run` method is calls `uploadchannel` (or `uploadchannel_wrapper`)
```
antonio-cortes-perez: remove "is"

ricecooker/sushi_bar_client.py
```diff
@@ -178,59 +241,51 @@ def report_statistics(self, files):
...
+
+class ControlWebSocket(ReconnectingWebSocket):
+    def __init__(self, chef, args, options):
+        self.chef = chef
+        self.cliargs = args
+        self.clioptions = options
+        self.channel = chef.get_channel()
+        if not self.channel:
+            raise SushiBarNotSupportedException(
+                'Chef does not implement create_channel')
```
antonio-cortes-perez: s/create_channel/get_channel/

ricecooker/sushi_bar_client.py
```diff
@@ -178,59 +241,51 @@ def report_statistics(self, files):
...
-        self.connect_lock.acquire()
-        self.stop_event.set()
-        self.ws.close()
-        self.connect_lock.release()
 
-    def stopped(self):
-        return self.stop_event.is_set()
 
 
 class LoggingHandler(logging.Handler):
```
antonio-cortes-perez: what do you think about moving this class to the monitoring section of the file?

ricecooker/chefs.py
```diff
@@ -0,0 +1,262 @@
...
+                title=self.channel_info['CHANNEL_TITLE'],
+                thumbnail=self.channel_info.get('THUMBNAIL'),
+            )
+            return channel
+
+        else:
+            raise NotImplementedError('BaseChef must overrride the get_channel method')
+
+
+    def construct_channel(self, **kwargs):
```
antonio-cortes-perez: One way to call get_channel only once, it is to receive here the channel node created. In this way, in uploadchannel, we will have something like:

channel = chef.get_channel(...)
...
channel = chef.construct_channel(channel, ...)

antonio-cortes-perez:
Also, could you please update the README.md Step 2?

ricecooker/chefs.py
ivanistheone: Good idea. I guess we can make it a keyword argument `channel=None` and if you don't pass in a channel we call  `get_channel` to get it. Another option is for the first call of  `get_channel` to also attach it's result to `self.channel`. Still doesn't seem any of these options are really necessary for now so will leave this for later... I'm sure ricecooker api will evolve and this is not an issue right now.

I'm thinking users can rely on the info in `self.channel_info` (dict). If they need extensibility, they can define a `get_channel_info` method that also returns a dict (this way they don't need to know what a `ChannelNode` is to provide the info).

The sushi bar integration code can get the `channel_id` using
`uuid.uuid5(uuid.uuid5(uuid.NAMESPACE_DNS, channel_info['CHANNEL_SOURCE_DOMAIN']), channel_info['CHANNEL_SOURCE_ID']).hex`

PR #96, opened 2017-07-05
Feature/newexample

ivanistheone:
Updated rest of chef scripts in the `examples/` directory to the new API.

### DIFF
--- a/examples/large_wikipedia_chef.py
+++ b/examples/large_wikipedia_chef.py
@@ -1,8 +1,11 @@
+#!/usr/bin/env python
+
 import tempfile
 
 import requests
 from bs4 import BeautifulSoup
 
+from ricecooker.chefs import SushiChef
 from ricecooker.classes import licenses
 from ricecooker.classes.files import HTMLZipFile
 from ricecooker.classes.nodes import ChannelNode, HTML5AppNode, TopicNode
@@ -13,7 +16,7 @@ from ricecooker.utils.zip import create_predictable_zip
 # CHANNEL SETTINGS
 SOURCE_DOMAIN = "<yourdomain.org>"  #
 SOURCE_ID = "<youasasrid>"  # an alphanumeric ID refering to this channel
-CHANNEL_TITLE = "<channeltitle>"  # a humand-readbale title
+CHANNEL_TITLE = "Large wikipedia chef"  # a humand-readbale title
 
 sess = requests.Session()
 cache = FileCache('.webcache')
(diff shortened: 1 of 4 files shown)


examples/large_wikipedia_chef.py
```diff
@@ -12,8 +15,8 @@
 
 # CHANNEL SETTINGS
 SOURCE_DOMAIN = "<yourdomain.org>"  #
-SOURCE_ID = "<yourid>"  # an alphanumeric ID refering to this channel
-CHANNEL_TITLE = "<channeltitle>"  # a humand-readbale title
+SOURCE_ID = "<youasasrid>"  # an alphanumeric ID refering to this channel
```
jamalex: Say what? :)

examples/large_wikipedia_chef.py
```diff
@@ -153,3 +169,12 @@ def process_wikipedia_page(content, baseurl, destpath, **kwargs):
...
 
     return str(page)
+
+
+
+if __name__ == '__main__':
+    """
+    This code will run when the sushi chef is called from the command line.
+    """
+    chef = LargeWikipdeiaChef()
```
jamalex: *LargeWikipediaChef

examples/tutorial_chef.py
```diff
@@ -0,0 +1,98 @@
...
+        # Now we are adding 'Example Subtopic' to our 'Example Topic'
+        exampletopic.add_child(examplesubtopic)
+        # TODO: Add your subtopic to your topic here
+
+
+        # Content
+        # You can add pdfs, videos, and audio files to your channel
+        ########################################################################
+        # let's create a document file called 'Example PDF'
+        examplepdf = DocumentNode(title="Example PDF", source_id="example-pdf", files=[DocumentFile(path="http://www.pdf995.com/samples/pdf.pdf")], license=licenses.CC_BY_SA)
```
jamalex: We support license constants for backwards-compatibility, but I think we have classes for them as well? May be better to use those in the demo, especially as I think the SpecialPermissions class takes an additional text argument, and so forth)

ivanistheone: Nice! I didn't know about the `get_license` "factory" function...

updating the PR with this

```
fancy_license = get_license(licenses.SPECIAL_PERMISSIONS, description='Special license for ricecooker fans only.', copyright_holder='The chef video makers')
```

produces:

screen shot 2017-07-05 at 9 13 15 pm

PR #97, opened 2017-07-05
KA assessment item util function.

ralphiee22:
Added a util function which recursively searches the Khan Academy assessment item for all url (key,value) pairs. These tend to be images embedded into widgets. This fixes missing images across all Khan Academy channels.

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -417,7 +417,7 @@ class YouTubeSubtitleFile(File):
             self.filename = self.download_subtitle()
             config.LOGGER.info("\t--- Downloaded subtitle {}".format(self.filename))
             return self.filename
-        except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


ivanistheone:
Looks good.

ricecooker/classes/files.py
```diff
@@ -417,7 +417,7 @@ def process_file(self):
             self.filename = self.download_subtitle()
             config.LOGGER.info("\t--- Downloaded subtitle {}".format(self.filename))
             return self.filename
-        except FileNotFoundError:
+        except (FileNotFoundError, youtube_dl.utils.DownloadError):
```
ralphiee22: I need to make a note that the reason I added this error was because there was some youtube ids that `youtube_dl` could not find. They have been removed from youtube. @ivanistheone

PR #98, opened 2017-07-07
SushiBar support for https:// and wss://

ivanistheone:
The old sushibar was: http://leq.sidewayspass.com/

The new sushibar is https://sushibar.learningequality.org/

This PR makes two small change to it chefs use SSL when reporing to the new bar, so that we can use this to run chefs with reporing:

```
export SUSHI_BAR_URL="sushibar.learningequality.org"
./mychef.py -v --token=... 
```


jamalex:
Great to have support for this, even better to have it as a default, but wondering whether it might be good to be able to override it when setting up a testbed locally?

ivanistheone:
Yes, that's a good point. We need a `dev` mode for local development and/or testing channels. We had previously discussed disabling some of the authentication/authorization checks as well in `dev` mode, so I'll add a discussion point for the triage meeting.

PR #101, opened 2017-08-09
Fixed sample program with construct_channel not being found

jayoshih:
Getting:
```
Traceback (most recent call last):
  File "c:\users\jordan\appdata\local\programs\python\python36\Lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\users\jordan\appdata\local\programs\python\python36\Lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\Jordan\ricecooker\ricecooker\__main__.py", line 9, in <module>
    chef.main()
  File "C:\Users\Jordan\ricecooker\ricecooker\chefs.py", line 199, in main
    self.run(args, options)
  File "C:\Users\Jordan\ricecooker\ricecooker\chefs.py", line 135, in run
    uploadchannel(self, **args_and_options)
  File "C:\Users\Jordan\ricecooker\ricecooker\commands.py", line 114, in uploadchannel
    channel = chef.construct_channel(**kwargs)
  File "C:\Users\Jordan\ricecooker\ricecooker\chefs.py", line 190, in construct_channel
    channel = self.chef_module.construct_channel(**kwargs)
AttributeError: module 'mod' has no attribute 'construct_channel'
```

Fixed by adding construct_channel method that calls chef's construct_channel method

### DIFF
--- a/examples/sample_program.py
+++ b/examples/sample_program.py
@@ -297,7 +297,7 @@ class SampleChef(SushiChef):
         'CHANNEL_DESCRIPTION': 'A sample sushi chef to demo content types.',      # (optional) description of the channel (optional)
     }
 
-    def construct_channel(self, **kwargs):
+    def construct_channel(self, *args, **kwargs):
         """
         Create ChannelNode and build topic tree.
         """
@@ -486,10 +486,13 @@ def parse_images(content):
     return content
 
 
-
+chef = SampleChef()
 if __name__ == '__main__':
     """
     This code will run when the sushi chef is called from the command line.
     """
-    chef = SampleChef()
+
     chef.main()
+
+def construct_channel(*args, **kwargs):
+    return chef.construct_channel(*args, **kwargs)
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -181,6 +181,7 @@ class File(object):
     filename = None
     language = None
     assessment_item = None
+    is_primary = False
 
     def __init__(self, preset=None, language=None, default_ext=None, source_url=None):
         self.preset = preset
@@ -295,6 +296,7 @@ class ThumbnailFile(ThumbnailPresetMixin, DownloadFile):
 class AudioFile(DownloadFile):
     default_ext = file_formats.MP3
     allowed_formats = [file_formats.MP3]
+    is_primary = True
 
     def get_preset(self):
         return self.preset or format_presets.AUDIO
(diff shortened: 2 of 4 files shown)


ivanistheone:
@jayoshih From the tracebackm it looks like you were calling the chef script using the "old style": `python -m ricecooker sample_program.py uploadchannel -v --token=...`

Will it work using the "new style" (as standalone script): `./sample_program.py -v --token=...` ?

examples/sample_program.py
```diff
@@ -297,7 +297,7 @@ class SampleChef(SushiChef):
         'CHANNEL_DESCRIPTION': 'A sample sushi chef to demo content types.',      # (optional) description of the channel (optional)
     }
 
-    def construct_channel(self, **kwargs):
+    def construct_channel(self, *args, **kwargs):
```
ivanistheone: Currently, the ricecooker doesn't pass any positional args to to `construct_channel` so it should work even without this, but I guess it's a good idea for scripts to also expect `*args` in case we add them in the future.

examples/sample_program.py
```diff
@@ -486,10 +486,13 @@ def parse_images(content):
     return content
 
 
-
+chef = SampleChef()
 if __name__ == '__main__':
     """
     This code will run when the sushi chef is called from the command line.
     """
-    chef = SampleChef()
```
ivanistheone: I think it makes more sense for the `chef` instance to be created inside the `__main__` block as before, but it should work either way.

examples/sample_program.py
```diff
@@ -486,10 +486,13 @@ def parse_images(content):
...
+
     chef.main()
+
+def construct_channel(*args, **kwargs):
```
ivanistheone: If you use the "new style" approach to calling the chef script, this shouldn't necessary.

ivanistheone:
Also, I just realized the "running as a standalone script" approach depends on the first line of the file, which specifies to run the script through Python  `#!/usr/bin/env python` (sometimes called a shebang line), which is a UNIX thing.

On Windows, you should call the script as `python sample_program.py -v --token=...`.

examples/sample_program.py
jayoshih: When I kept it in the block, it said chef didn't exist

examples/sample_program.py
jayoshih: Getting an error with the get_channel method looking for *args

PR #102, opened 2017-08-11
[WIP] smart YouTubeSubtitleFile language detection

ivanistheone:
Followup from language code discussions.

WIP because `languages.getlang_by_alpha2` doesn't exist yet, but will exist soon...

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -404,8 +404,21 @@ class YouTubeVideoFile(WebVideoFile):
         super(YouTubeVideoFile, self).__init__('http://www.youtube.com/watch?v={}'.format(youtube_id), **kwargs)
 
 class YouTubeSubtitleFile(File):
+    """
+    Helper class for downloading youtube subtitles.
+    Args:
+       youtube_id (string): YouTube ID of video
+       language (string): internal language id format `{primary_code}` or `{primary_code}-{subcode}` (required)
+                          alternatively, can proviede two-letter language code recognized by youtube
+    """
     def __init__(self, youtube_id, language=None, **kwargs):
         self.youtube_url = 'http://www.youtube.com/watch?v={}'.format(youtube_id)
+        self.youtube_language = language  # two-letter code, sometimes different from internal repr.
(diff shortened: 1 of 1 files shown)


ricecooker/classes/files.py
```diff
@@ -404,8 +404,21 @@ def __init__(self, youtube_id, **kwargs):
         super(YouTubeVideoFile, self).__init__('http://www.youtube.com/watch?v={}'.format(youtube_id), **kwargs)
 
 class YouTubeSubtitleFile(File):
+    """
+    Helper class for downloading youtube subtitles.
+    Args:
+       youtube_id (string): YouTube ID of video
+       language (string): internal language id format `{primary_code}` or `{primary_code}-{subcode}` (required)
+                          alternatively, can proviede two-letter language code recognized by youtube
```
divad12: nit: provide

ricecooker/classes/files.py
```diff
@@ -404,8 +404,21 @@ def __init__(self, youtube_id, **kwargs):
...
+    """
     def __init__(self, youtube_id, language=None, **kwargs):
         self.youtube_url = 'http://www.youtube.com/watch?v={}'.format(youtube_id)
+        self.youtube_language = language  # two-letter code, sometimes different from internal repr.
```
divad12: nit: could actually be something like `zh-Hans` too

ricecooker/classes/files.py
```diff
@@ -404,8 +404,21 @@ def __init__(self, youtube_id, **kwargs):
...
+        language_obj = languages.getlang(language)   # lookup `language` using internal representation
+        # if language_obj not None, we know `language` is a valid language_id in internal repr.
+        if language_obj is None:  # if `language` not found using internal repr.
+            language_obj = languages.getlang_by_alpha2(language)  # try to match by two-letter ISO code
+            language = language_obj.id  # update `language` argument from internal repr. language_id
```
divad12: How about adding a `None`-guard here:

```
if language_obj:
    language = language_obj.id
```

... and letting the error propagate to the superclass, to retain the old behaviour for failure-to-match language.

PR #103, opened 2017-08-11
Feature/node language

ivanistheone:
Can now specify `language` (internal repr. code or Language object)  for `Node` base class.

This means ChennalNode, TopicNode, and all subclasses of ContentNode now support setting language for them.

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -189,13 +189,15 @@ class File(object):
         self.source_url = source_url
 
     def set_language(self, language):
-        self.language = language
-        if isinstance(self.language, str):
-            self.language = languages.getlang(language)
-            if not self.language:
-                raise TypeError("Language {} is not found".format(language))
-        if isinstance(self.language, languages.Language):
-            self.language = self.language.code
+        """ Set self.language to internal lang. repr. code from str or Language object. """
+        if isinstance(language, str):
+            language_obj = languages.getlang(language)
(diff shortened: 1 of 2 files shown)


ricecooker/classes/files.py
```diff
@@ -189,13 +189,15 @@ def __init__(self, preset=None, language=None, default_ext=None, source_url=None
...
-        self.language = language
-        if isinstance(self.language, str):
-            self.language = languages.getlang(language)
-            if not self.language:
-                raise TypeError("Language {} is not found".format(language))
-        if isinstance(self.language, languages.Language):
-            self.language = self.language.code
+        """ Set self.language to internal lang. repr. code from str or Language object. """
+        if isinstance(language, str):
+            language_obj = languages.getlang(language)
```
jayoshih: Alternatively, you could set this to be `language` instead of `language_obj` then set self.language under the `if isinstance(language, languages.Language)` block (not important at all, but could make code more concise)

ricecooker/classes/nodes.py
```diff
@@ -217,6 +230,7 @@ def to_dict(self):
             "id": self.get_node_id().hex,
             "name": self.title,
             "thumbnail": self.thumbnail.filename if self.thumbnail else None,
+            "language" : self.language,
```
jayoshih: CC is going to look for "language_id" to make it clearer that it's the id, not the language object

PR #109, opened 2017-08-28
Tutorial notebook for setting node languages

ivanistheone:
Additional docs about languages in notebook form.


ivanistheone:
Can view rendered notebook here: https://github.com/ivanistheone/ricecooker/blob/feature/languages_notebook/docs/tutorial/languages.ipynb

PR #121, opened 2017-09-25
Jsonchef, LineCook and supporting utils

ivanistheone:
JsonTreeChef and LineCook code

### DIFF
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -742,8 +742,6 @@ class ExerciseNode(ContentNode):
             assert self.kind == content_kinds.EXERCISE, "Assumption Failed: Node should be an exercise"
 
             # Check if questions are correct
-            if not any(self.questions):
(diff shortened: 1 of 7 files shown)


ivanistheone:
For context, the idea behind the `LineCook` sushi chef is so that the code for souschef-based channels can be very simple, e.g., https://github.com/learningequality/sushi-chef-openstax/blob/master/sushichef.py

ricecooker/classes/nodes.py
```diff
@@ -742,6 +742,8 @@ def validate(self):
             assert self.kind == content_kinds.EXERCISE, "Assumption Failed: Node should be an exercise"
 
             # Check if questions are correct
+            if not any(self.questions):
```
jayoshih: Might be better to combine this with the assertion error ("Assumption Failed: Exercise does not have a question ({})".format(self.__dict__))

PR #122, opened 2017-10-02
 Added validation on source ids and copyright holders

jayoshih:


tests/test_data.py
```diff
@@ -369,7 +369,6 @@ def test_init(channel, topic, video, audio, document, html, exercise):
 
 def test_validate(channel, invalid_channel, topic, contentnode_invalid_license, contentnode_invalid_files, contentnode_no_source_id, video, video_invalid_files,
     audio, audio_invalid_files, document, document_invalid_files, html, html_invalid_files, html_invalid_zip, exercise, exercise_invalid_question):
-    assert channel.validate(), "Valid channel should pass validation"
```
jayoshih: Note: Had to remove this as all of the children have the same source id. Fixing this would cause the rest of the tests to fail

ivanistheone: Yes, I saw this. I'm working on refactoring these tests around this helper function:

```
def genrate_random_ids(channel_domain_namespace, channel_node_id):
    source_id = ''.join(random.choices(string.ascii_uppercase + string.digits, k=10))
    content_id = uuid.uuid5(channel_domain_namespace, source_id)
    node_id = uuid.uuid5(channel_node_id, content_id.hex)
    ids_dict = dict(
        source_id=source_id,
        content_id=content_id.hex,
        node_id=node_id.hex,
    )
    return ids_dict
```

it's a bit of work but I think we want to keep the channel validate test as it's super crucial...

ivanistheone:
Cherry picked the commits to #123 and merged.

PR #129, opened 2017-10-31
Feature/cmdsock

ivanistheone:
Copy-pasta from WebSocketControl class, but seems to work fine.

TODO: learn more about Python threads and make sure all threads are OK.

TODO: setup a test cronjob on `vader` to make sure this works before sharing with other cheffist

### DIFF
--- a/ricecooker/sushi_bar_client.py
+++ b/ricecooker/sushi_bar_client.py
@@ -5,7 +5,9 @@ import requests
 import subprocess
 import threading
 import time
+import socket
 import websocket
+
 
 from . import config
 from . import __version__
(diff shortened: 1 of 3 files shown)


ricecooker/sushi_bar_client.py
```diff
@@ -59,7 +61,7 @@ def send(self, data):
                 self.ws.send(data)
                 return
             except websocket.WebSocketConnectionClosedException:
-                print('WebSocket closed, retrying send.')
+                config.LOGGER.debug('WebSocket closed, retrying send.')
```
jayoshih: Were you able to get it to run locally with this? I tried doing this, and it caused some sort of infinite loop (might be resolved now, but not sure)

ivanistheone: Haven't tried localhost no... I will keep an eye out for this before release to PyPI.

Maybe it's trying to log to the socket, and then remote-web-socket is failing which generates a new log message etc. I can see how this could be bad...

PR #130, opened 2017-11-14
Moved utils from cookiecutter to ricecooker

jayoshih:


ivanistheone:
See #131 for same changes as above + long story featuring lots of git commands.

PR #132, opened 2017-11-22
Fully close selenium webdriver's spawned processes

divad12:
Previously, after each run of the Open Osmosis sushi chef, there would be remaining `phantomjs` processes still running, which would clog up the network bandwidth quite a bit. Apparently calling `driver.close()` doesn't suffice to fully close out all spawned processes of PhantomJS, as documented in https://github.com/seleniumhq/selenium/issues/767 . This fixes that.

Tested by running the Open Osmosis sushi chef pointing to the local checkout of Ricecooker with these changes (`pip install -e /path/to/local/ricecooker/`) and ensuring there are no `phantomjs` processes left over after the run (whereas there were before this change).

### DIFF
--- a/ricecooker/utils/html.py
+++ b/ricecooker/utils/html.py
@@ -1,5 +1,6 @@
 import os
 import requests
+import signal
 import time
 import urllib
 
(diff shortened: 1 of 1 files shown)


divad12:
Review please @ivanistheone  or @jamalex ?

ivanistheone:
Looks good; merging.

@jayoshih Please test this works as expected on Windows as well

PR #146, opened 2018-02-15
Fix problem of svg files not being downloaded due to missing protocol

ralphiee22:
Graphie files wouldn't be properly generated because it would try to download from url like `//ka-perseus-graphie.s3.amazonaws.com/883480314c702c447570bcee34b2b435d3149ebf.svg` instead of `http://ka-perseus-graphie.s3.amazonaws.com/883480314c702c447570bcee34b2b435d3149ebf.svg`


ivanistheone:
Looks good. This fix will be in the next ricecooker release, which is probably this Friday. 
@ralphiee22 Let me know if you need it sooner.

PR #147, opened 2018-02-18
Ricecooker ePub support [WIP]

ivanistheone:
Minimal changes needed to implement ePub support.

This also PR contains detailed instructions for testing associated changes in 
  - le-utils https://github.com/learningequality/le-utils/pull/39
  - Studio https://github.com/fle-internal/content-curation/pull/530
  - Kolibri  https://github.com/learningequality/kolibri/pull/3210 

TODO:
  - review assumptions: 1) New file type, new file_format, new file preset, but reuse `document` content kind
  - update JsonChef + CSV chef workflows


ricecooker/classes/files.py
```diff
@@ -356,14 +358,32 @@ class AudioFile(DownloadFile):
...
     default_ext = file_formats.PDF
-    allowed_formats = [file_formats.PDF]
+    allowed_formats = [file_formats.PDF, file_formats.EPUB]
     is_primary = True
 
     def get_preset(self):
         return self.preset or format_presets.DOCUMENT
 
+
+class EPubFile(DownloadFile):
```
jayoshih: When would one use the DocumentFile vs the EPubFile? Looks like epub is accepted under the document file, but wondering if it would make sense to just keep these separate entirely to avoid any confusion

ricecooker/classes/files.py
```diff
@@ -356,14 +358,32 @@ class AudioFile(DownloadFile):
...
+    default_ext = file_formats.EPUB
+    allowed_formats = [file_formats.EPUB]  # ZEPUB other ???
+    is_primary = True
+
+    def get_preset(self):
+        return self.preset or format_presets.EPUB
+
+    def process_file(self):
```
jayoshih: Was something going to be added here? If not, do we still need to implement the function here? (same with the validate function)

Issue #134, opened 2017-11-28
Buried type error in exception case in sushi_bar_client


dragondave:
Whilst trying to get a ricecooker up and running, I hit a chain of issues which might be worth fixing.

https://github.com/learningequality/ricecooker/blob/fb625eca06b1e6f803c4a8f7d6c64c3183f1625d/ricecooker/sushi_bar_client.py#L91

assumes `.__config_logger()` returns two items. But it returns `None` if `.run_id` is falsey.

https://github.com/learningequality/ricecooker/blob/fb625eca06b1e6f803c4a8f7d6c64c3183f1625d/ricecooker/sushi_bar_client.py#L160-L162

`.run_id` is the return value of `.__create_channel_run`

https://github.com/learningequality/ricecooker/blob/fb625eca06b1e6f803c4a8f7d6c64c3183f1625d/ricecooker/sushi_bar_client.py#L89

which has a broad sweeping exception which will silence the exception and return None.

(The exception in this case was a 400 for the post  'https://sushibar.learningequality.org/api/channelruns' -- will dig deeper separately; suspect API key wrong.)

I suggest `.__create_channel_run` possibly shouldn't silence the exception at all since it's probably impossible to recover gracefully from, and it should just crash ASAP to avoid later type errors. (It might be useful elsewhere).

dragondave:
[it would be useful to `print (request.text)`, then `raise` -- turns out `chef_name` was a mandatory field]

dragondave:
The same logic possibly applies to `__get_chef_name`: the root cause appears to be the git remote URL not being set up (!)

ivanistheone:
Looking into this now, and posting the logs of the error that occur when running a good chef outside a git repo:

```
Chef name: Command '['git', 'config', '--get', 'remote.origin.url']' returned non-zero exit status 1.
Error channel run: 400 Client Error: Bad Request for url: https://sushibar.learningequality.org/api/channelruns/
run_id: None
'NoneType' object is not iterable
Traceback (most recent call last):
  File "./chef.py", line 31, in <module>
    chef.main()
  File "/private/tmp/cheftest/venv/lib/python3.6/site-packages/ricecooker/chefs.py", line 293, in main
    self.run(args, options)
  File "/private/tmp/cheftest/venv/lib/python3.6/site-packages/ricecooker/chefs.py", line 284, in run
    uploadchannel_wrapper(self, args, options)
[...]

PR #149, opened 2018-02-28
Bugfix for non-git chef runs

ivanistheone:
In this PR:
  - fix for https://github.com/learningequality/ricecooker/issues/134
  - Rename contentcuration URLs to studio URLs
     - Add support for new environment variable STUDIO_URL (in preparation for upcoming docs update)
  - Enabled `--nomonitor` command line argument (if you want to do a chef run that doesn't talk to sushibar, e.g. for testing purposes etc.)

### DIFF
--- a/ricecooker/config.py
+++ b/ricecooker/config.py
@@ -15,8 +15,11 @@ SUSHI_BAR_CLIENT = None
 STAGE = False
 LOGGER = logging.getLogger()
 
-# Domain and file store location for uploading to production server
-DOMAIN = os.getenv('CONTENTWORKSHOP_URL', "https://contentworkshop.learningequality.org")
+# Domain and file store location for uploading to production Studio server
+DOMAIN_ENV = os.getenv('STUDIO_URL', None)
+if DOMAIN_ENV is None:  # check old ENV varable for backward compatibility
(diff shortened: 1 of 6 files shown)


ricecooker/config.py
```diff
@@ -15,8 +15,11 @@
 STAGE = False
 LOGGER = logging.getLogger()
 
-# Domain and file store location for uploading to production server
-DOMAIN = os.getenv('CONTENTWORKSHOP_URL', "https://contentworkshop.learningequality.org")
+# Domain and file store location for uploading to production Studio server
+DOMAIN_ENV = os.getenv('STUDIO_URL', None)
+if DOMAIN_ENV is None:  # check old ENV varable for backward compatibility
+    DOMAIN_ENV = os.getenv('CONTENTWORKSHOP_URL', None)
+DOMAIN = DOMAIN_ENV if DOMAIN_ENV else "https://studio.learningequality.org"
```
jayoshih: Might want to keep contentworkshop since cloudflare prevents uploads of over 500MB

ivanistheone: @jayoshih 500MB should be OK, but since you were asking about a possible workaround in case we want to support 500MB+ single-file uploads, the ricecooker user could set SUSHIBAR_URL env var to "https://contentworkshop.learningequality.org" to avoid the cloudflare middleman.

PR #150, opened 2018-03-02
Add a function for downloading static assets for an HTML page

divad12:
Tested with the Blockly, 3asafeer, and Thoughtful Learning sushi chefs. Corresponding PRs to use this function for each of the above will follow.

### DIFF
--- a/ricecooker/utils/downloader.py
+++ b/ricecooker/utils/downloader.py
@@ -1,8 +1,15 @@
+import os
+import re
 import requests
 import time
+from urllib.parse import urlparse, parse_qs
+import uuid
+
+from bs4 import BeautifulSoup
 from selenium import webdriver
 from requests_file import FileAdapter
 from ricecooker.config import PHANTOMJS_PATH
+from ricecooker.utils.html import download_file, WebDriver
 from ricecooker.utils.caching import CacheForeverHeuristic, FileCache, CacheControlAdapter, InvalidatingCacheControlAdapter
 
 DOWNLOAD_SESSION = requests.Session()                          # Session for downloading content from urls
@@ -13,6 +20,12 @@ forever_adapter= CacheControlAdapter(heuristic=CacheForeverHeuristic(), cache=ca
 
 DOWNLOAD_SESSION.mount('http://', forever_adapter)
 DOWNLOAD_SESSION.mount('https://', forever_adapter)
+
+headers = {
+    "User-Agent": "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:20.0) Gecko/20100101 Firefox/20.0",
+    "Accept-Encoding": "gzip, deflate",
+    "Connection": "keep-alive"
+}
 
 
 def read(path, loadjs=False, session=None, driver=None):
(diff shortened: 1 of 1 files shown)


ricecooker/utils/downloader.py
```diff
@@ -41,3 +54,180 @@ def read(path, loadjs=False, session=None, driver=None):
...
+        try:
+            response = sess.get(url, headers=headers, timeout=timeout, *args, **kwargs)
+            break
+        except (requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout) as e:
+            retry_count += 1
+            print("Error with connection ('{msg}'); about to perform retry {count} of {trymax}."
+                  .format(msg=str(e), count=retry_count, trymax=max_retries))
+            time.sleep(retry_count * 1)
+            if retry_count >= max_retries:
+                return Dummy404ResponseObject(url=url)
```
ivanistheone: The `Dummy404ResponseObject` is missing.

It is defined here:  https://github.com/learningequality/sushi-chef-ck12/blob/master/chef.py#L77
but may be better to refactor to return `None` (or some other way to signal error).

ricecooker/utils/downloader.py
```diff
@@ -41,3 +54,180 @@ def read(path, loadjs=False, session=None, driver=None):
...
+
+
+# TODO(davidhu): Use MD5 hash of URL (ideally file) instead.
+def _derive_filename(url):
+    name = os.path.basename(urlparse(url).path).replace('%', '_')
+    return ("%s.%s" % (uuid.uuid4().hex, name)).lower()
+
+
+def make_fully_qualified_url(url, base):
+    # TODO(davidhu): This is some hacky special-casing for 3asafeer ...
```
ivanistheone: I believe some of these can be handled by `urllib.parse.urljoin`, e.g., 
```
from urllib.parse import urljoin

base = 'http://somesite.com/subpath/file.html'

print(urljoin(base, '../image'))
http://somesite.com/image

print(urljoin(base, '../scripts'))
http://somesite.com/scripts

print(urljoin(base, '//some.full.path/just/no/scheme.html'))
http://some.full.path/just/no/scheme.html

print(urljoin(base, '/an/abs/path.html'))
http://somesite.com/an/abs/path.html

print(urljoin(base, 'a/rel/path.html'))
http://somesite.com/subpath/a/rel/path.html
```

ricecooker/utils/downloader.py
```diff
@@ -1,8 +1,15 @@
...
 import requests
 import time
+from urllib.parse import urlparse, parse_qs
+import uuid
+
+from bs4 import BeautifulSoup
 from selenium import webdriver
 from requests_file import FileAdapter
 from ricecooker.config import PHANTOMJS_PATH
+from ricecooker.utils.html import download_file, WebDriver
```
ivanistheone: `WebDriver` is not used

divad12: Removed

ricecooker/utils/downloader.py
divad12: Changed to `raise e`

ricecooker/utils/downloader.py
divad12: This is _awesome_, thanks for sharing!!

divad12:
Thanks for the review @ivanistheone !

> Also, not sure if we can combine the new function with the read method from the other PR? Or maybe keep thing separated.

I think it'd be best to keep things separated, to give users the flexibility of getting a page's HTML without also grabbing all of the static assets, e.g. if we're scraping a video page and want to fetch metadata from the HTML source. Thanks for suggesting though!

PR #152, opened 2018-03-03
Improve downloader.read with retry logic and (hopefully) reduced JS waiting

divad12:
(Review is not urgent)

1. For the non-loading-JS case, add some retry-and-wait logic, copied from previous chefs.
2. For the load-JS case, parameterize the wait time and change its default from 5 to 3 seconds, since [Selenium Phantom JS already waits for the page to load](https://stackoverflow.com/questions/26566799/how-to-wait-until-the-page-is-loaded-with-selenium-for-python/26567563#26567563) in the call to `driver.get()`. (So the wait time would just be time for the JS to execute and make any changes to the page before we grab its source.) Also, added a callback parameter to more dynamically tell us when we can grab the page source.

Test plan:

1. 

```
import ricecooker.utils.downloader as downloader

source = downloader.read(
    "http://3asafeer.com/",
    loadjs=False,
)
print('Page source is:', source)
```

2.

```
import ricecooker.utils.downloader as downloader

source = downloader.read(
    "http://3asafeer.com/",
    loadjs=True,
    loadjs_wait_time=1,
    loadjs_wait_for_callback=lambda driver: driver.find_element_by_id('readLink')
)
print('Page source is:', source)
```

### DIFF
--- a/ricecooker/utils/downloader.py
+++ b/ricecooker/utils/downloader.py
@@ -1,6 +1,7 @@
 import requests
 import time
 from selenium import webdriver
+import selenium.webdriver.support.ui as selenium_ui
 from requests_file import FileAdapter
 from ricecooker.config import PHANTOMJS_PATH
 from ricecooker.utils.caching import CacheForeverHeuristic, FileCache, CacheControlAdapter, InvalidatingCacheControlAdapter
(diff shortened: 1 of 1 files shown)


ivanistheone:
PR looks good. I'll add the examples you posted above to the docs when I update next.

PR #154, opened 2018-03-15
Updates to downloader.download_static_assets for MEET chef

divad12:
- make `derive_filename` an argument
- skip downloading URLs that start with `data:`
- fix a bug to actually use the `request_fn` from passed in argument

Supports PR https://github.com/learningequality/sushi-chef-MEET/pull/1


ivanistheone:
Looking good.

PR #156, opened 2018-03-28
Added role to node

jayoshih:


ivanistheone:
Looks good. Since the default value is set in `Node.__init__`, this should make it backward compatible, so no need to update chefs.

PR #166, opened 2018-05-18
No matching language code defaults to english

ralphiee22:
There have been some cases where we do not have a youtube subtitle language within our internal representations (Specifically when running the KA english chef). This causes the sushi chef to error out.

I propose we have an `unknown` lang code. Or maybe we even allow the subtitles to be added but with the youtube subtitle defined language code.


ivanistheone:
alternative check-first approach implemented in https://github.com/learningequality/ricecooker/pull/168

PR #172, opened 2018-06-12
Ensure web+graphie protocol is on KA questions

ralphiee22:
We need the `web+graphie:` protocol in order for perseus renderer to locate the correct `.svg` and `-data.json` files.

### DIFF
--- a/ricecooker/classes/questions.py
+++ b/ricecooker/classes/questions.py
@@ -181,7 +181,7 @@ class BaseQuestion:
         filename = exercise_file.process_file()
 
         # Need to replace text's web+graphie with https to get text matches
-        text = text.replace("web+graphie://", "https://").replace(path_text, exercises.CONTENT_STORAGE_FORMAT.format(exercise_file.get_replacement_str()))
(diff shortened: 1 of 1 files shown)


ivanistheone:
@jayoshih  recently made this related change https://github.com/learningequality/ricecooker/commit/dbcf33e4986ca858e221548fae2e090a43447604#diff-75507a40a038eac74e3670dd2a70a781R184   so I think she'll be best to review this.

ricecooker/classes/questions.py
```diff
@@ -181,7 +181,7 @@ def set_image(self, text):
         filename = exercise_file.process_file()
 
         # Need to replace text's web+graphie with https to get text matches
-        text = text.replace("web+graphie://", "https://").replace(path_text, exercises.CONTENT_STORAGE_FORMAT.format(exercise_file.get_replacement_str()))
+        text.replace("web+graphie://", "web+graphie:https://").replace(path_text, exercises.CONTENT_STORAGE_FORMAT.format(exercise_file.get_replacement_str()))
```
jayoshih: Might need to change it to "text = text.replace..." otherwise the text won't get set to the new value

ralphiee22: whoops!

PR #174, opened 2018-06-25
[WIP] Serving up iframes

kollivier:
This PR makes the local server browser preview feature callable from the command line, and adds a special iFrame mode that lets you run tests from within a iFrame sandbox mimicking Kolibri's.

### DIFF
--- a/preview.py
+++ b/preview.py
@@ -1,0 +1,21 @@
+import sys
+
+import ricecooker.utils.browser
+
+if __name__ == '__main__':
+
+    iframe = False
+
+    if "--iframe" in sys.argv:
+        sys.argv.pop(sys.argv.index("--iframe"))
+        iframe = True
+
+    get_help = "-h" in sys.argv or "--help" in sys.argv
+    if get_help or len(sys.argv) < 2:
+        print("usage: preview.py [--iframe] /path/to/page/or/directory")
+        print("Preview web content in a browser using a local server.")
+        print("Options:")
+        print("--iframe:    load the content into an iframe sandbox for security testing.")
+        print("-h, --help:  display this help message.")
+        sys.exit(int(not get_help))
+    ricecooker.utils.browser.load_server(sys.argv[1], iframe)
--- a/ricecooker/utils/browser.py
+++ b/ricecooker/utils/browser.py
@@ -1,8 +1,62 @@
-import os, urllib, posixpath, webbrowser
-from http.server import HTTPServer, BaseHTTPRequestHandler, SimpleHTTPRequestHandler
+import os
+import posixpath
+import sys
+import urllib
+import webbrowser
+
+from http.server import HTTPServer, HTTPStatus, SimpleHTTPRequestHandler
+
+SERVER = "127.0.0.1"
+PORT = 8282
 
 
-def preview_in_browser(directory, filename="index.html", port=8282):
+class IFrameServerRequestHandler(SimpleHTTPRequestHandler):
+    def do_OPTIONS(self):
+        self.send_access_control_headers()
+        self.send_response(200, "ok")
+
+    def send_access_control_headers(self):
+        self.send_header('Access-Control-Allow-Origin', '*')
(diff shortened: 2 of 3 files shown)


preview.py
```diff
@@ -0,0 +1,21 @@
+import sys
```
jamalex: Would it perhaps be good to have this (or something with a distinctive enough name) added to the path after you pip install ricecooker? Maybe a `ricecooker` command with subcommands like `ricecooker preview ...`? Since most people aren't cloning the source, it's otherwise a pain to find this file in the installed packages etc in order to run it.

ricecooker/utils/browser.py
```diff
@@ -1,8 +1,62 @@
...
+        except OSError:
+            self.send_error(HTTPStatus.NOT_FOUND, "File not found")
+            return None
+        try:
+            self.send_response(HTTPStatus.OK)
+            self.send_header("Content-type", ctype)
+            fs = os.fstat(f.fileno())
+            self.send_header("Content-Length", str(fs[6]))
+            self.send_header("Last-Modified", self.date_time_string(fs.st_mtime))
+            self.send_header("Cache-Control", "no-cache, no-store, must-revalidate")
```
jamalex: nice!

ricecooker/utils/browser.py
```diff
@@ -1,8 +1,62 @@
...
+        try:
+            f = open(path, 'rb')
+        except OSError:
+            self.send_error(HTTPStatus.NOT_FOUND, "File not found")
+            return None
+        try:
+            self.send_response(HTTPStatus.OK)
+            self.send_header("Content-type", ctype)
+            fs = os.fstat(f.fileno())
+            self.send_header("Content-Length", str(fs[6]))
```
jamalex: Isn't there an attribute for the file size on `fs`? Might be clearer.

ricecooker/utils/browser.py
```diff
@@ -21,8 +75,32 @@ def translate_path(self, path):
...
+
+    webbrowser.open("http://{}:{}/{}".format(SERVER, PORT, filename))
+
+    httpd.serve_forever()
 
-    webbrowser.open("http://127.0.0.1:{}/{}".format(port, filename))
+def load_server(path, iframe=False):
+    if iframe:
+        print("Starting local server and iframe sandbox...")
+        iframe_preview(path)
```
jamalex: I missed where the wrapper page gets served up, with the iframe itself, and its sandbox attribute?

preview.py
kollivier: I see this approach as more of a band-aid for internal use than a solution we want to commit to, so I didn't go too far with it. I think the ultimate fix for this will be two-fold:

1. For content developers: A working Studio -> Kolibri preview / test workflow, maybe from dev boxes so we don't affect Kolibri proper.
2. For Kolibri / Studio devs: Working integration tests, so that we can not only test that it works, but that it keeps working day in and day out, like the Energizer Bunny. Even when you update Chrome. ;)

I'd only want to really commit to this route and polish it into something easy and accessible once I know that #2 is not really viable, but right now I still think there's a solution to be had there. This just helps us to keep making progress on the HTML5 front until we get there.

ricecooker/utils/browser.py
kollivier: Right now the path is to the iframe itself, not the embedded content. Currently, to use this for testing, the code in the iframe itself will have to change depending on the test. (e.g. to check that the iframe parent received the messages for the test in the right order, etc.) See the packages/hashi/test directory on the https://github.com/learningequality/kolibri/tree/html5_app_bridge_library branch for a couple examples.

ricecooker/utils/browser.py
kollivier: Yeah, I just pushed a change for this, thanks!

kollivier:
Closing pull request for now as we can now upload and preview changes easily in Studio.

PR #181, opened 2018-08-30
Switch default url to api.studio.learningequality.org

jayoshih:


ivanistheone:
Looks good to me. From what I see from DNS, this is another GCP instance and doesn't pass through the CDN. Good idea to separate the chef-backend-service from the web-service.

Issue #185, opened 2018-09-12
create_predictable_zip does not close the temp file


lyw07:
* ricecooker version: 0.6.22
* Python version: 3.6
* Operating System: Mac OSX

### Description

When using `create_predictable_zip` to create zip files for HTML5Apps, the temp file created in the function `create_predictable_zip` is not closed after use. 
So after creating many zip files, the program fails with the error below

### What I Did
```
[Errno 24] Too many open files: '/var/folders/w6/97k87rrd7z30njcc74049z800000gn/T/tmpn5zdtpbz.zip'
Traceback (most recent call last):
  File "sushichef.py", line 223, in <module>
    chef.main()
  File "/Users/lingyiwang/.venvs/sushichef-pd/lib/python3.6/site-packages/ricecooker/chefs.py", line 296, in main
    self.run(args, options)
  File "/Users/lingyiwang/.venvs/sushichef-pd/lib/python3.6/site-packages/ricecooker/chefs.py", line 287, in run
    uploadchannel_wrapper(self, args, options)
  File "/Users/lingyiwang/.venvs/sushichef-pd/lib/python3.6/site-packages/ricecooker/commands.py", line 33, in uploadchannel_wrapper
    uploadchannel(chef, **args_and_options)
  File "/Users/lingyiwang/.venvs/sushichef-pd/lib/python3.6/site-packages/ricecooker/commands.py", line 130, in uploadchannel
    channel = chef.construct_channel(**kwargs)
  File "sushichef.py", line 105, in construct_channel
    self.download_subject(topic[0], topic[1])
  File "sushichef.py", line 168, in download_subject
    self.download_content(age_topic, link, page_params, selected_category, i*20)
  File "sushichef.py", line 203, in download_content
    zip_path = create_predictable_zip(filepath)
  File "/Users/lingyiwang/.venvs/sushichef-pd/lib/python3.6/site-packages/ricecooker/utils/zip.py", line 33, in create_predictable_zip
    with zipfile.ZipFile(zippath, "w") as outputzip:
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/zipfile.py", line 1090, in __init__
    self.fp = io.open(file, filemode)
OSError: [Errno 24] Too many open files: '/var/folders/w6/97k87rrd7z30njcc74049z800000gn/T/tmpn5zdtpbz.zip'
```

ivanistheone:
[...]

PR #186, opened 2018-09-13
Fix/predictable zip

ivanistheone:
This is a fix for  https://github.com/learningequality/ricecooker/issues/185

The issue was a little tricky detail --- the call to `tempfile.mkstemp(suffix=".zip")` opens a file descriptor `fd` (int) for this file and it's on us to close it when we're done.

Found via
https://ubuntuforums.org/showthread.php?t=919340&s=a051814d3439ffe92a8edb3edd7b583b&p=5786446#post5786446

I assume the remporary zip file will stick around after the fd is closed... it better because ricecooker will look for the zip file at this path later on.

### DIFF
--- a/ricecooker/utils/zip.py
+++ b/ricecooker/utils/zip.py
@@ -7,12 +7,13 @@ def _read_file(path):
         return f.read()
 
 def create_predictable_zip(path):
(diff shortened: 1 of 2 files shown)


jayoshih:
Nice catch!

PR #204, opened 2019-03-25
Add transcoding and node creation utilties and utility documentation

dragondave:
Adding a few utilites that I've created to make cheffing easier. Currently getting tests up and running and checking things over.


kollivier:
Cool, thanks! What do you think about switching to generating API docs instead of putting everything in the README? I tend to prefer when the docs are available both in online docs and in the code itself. It would also encourage us to document all the Ricecooker functions.

dragondave:
I think the biggest problem with autogenerated docs at this point is that there's a couple of very different audiences -- people hacking on ricecooker's own internals, and the public [python, not rest] API that's anticipated to be used by cheffers. I think I'd prefer to widen that gap -- my original intention was to package up these functions as their own thing, as they're rather ancilliary to ricecooker.

I notice we've already got a readthedocs set up, but it's very barebones at the moment.
https://learningequalityricecooker.readthedocs.io/en/latest/#

Having better API docs is a good thing, and I am generally in favour of autogeneration. I'll take a look at what it'd take to get readthedocs / sphinx properly set up, but I don't see a reason to let that block this PR.

dragondave:
(In hindsight, I probably shouldn't have `git amend`ed just now, but everything seems OK.)

dragondave:
I'm torn as to whether to update `ricecooker/utils/__init__.py` with short names for the functions I'd want to be calling -- the primary reason not to in my opinion is that some of the files have setup code which might be expensive and it would be inappropriate to invoke every time, especially if the functions themselves aren't being invoked.

(It's possible I'm misinterpretting how Python imports things: I'm going to take a look at how hard this would be)

kollivier:
Sorry for the late reply, we spent the last week auditing our content dev processes so we've been busy with that!

I think the docs should either go in the functions themselves if they're function reference docs, or more walkthrough-style docs should go in ricecooker/docs, which is what is published on pypi right now. My main concern here is consistency, so we have clear rules for what docs go where.

Regarding imports, in Python typically the convention is to either import the module with a short name, or import the function directly from the module, e.g. 'from ricecooker.utils.add_file import create_node` or `from ricecooker.utils import add_file; add_file.create_node(...)`. Directly importing sub-modules into `__init__.py` generally has undesirable side-effects like longer import times and a polluted namespace.

I'll post more detailed comments about the code shortly!

ricecooker/utils/add_file.py
```diff
@@ -0,0 +1,204 @@
...
+        new_filename = filename + extension
+        os.rename(filename, new_filename)
+        filename = new_filename
+
+    # print (filename, os.path.getsize(filename))
+
+    # Do not permit zero-byte files
+    assert(os.path.getsize(filename))
+
+    kwargs = {VideoFile: {"ffmpeg_settings": {"max_width": 480, "crf": 28}},
```
kollivier: In the case of TranscodedVideo, won't this cause the video to be converted a second time? Maybe we could integrate the conversion / transcode features into pressurecooker? e.g. tweak the functions here and maybe add an audio-only mode? https://github.com/learningequality/pressurecooker/blob/master/pressurecooker/videos.py

ricecooker/utils/add_file.py
```diff
@@ -0,0 +1,204 @@
...
+
+node_dict = {VideoFile: VideoNode,
+             AudioFile: AudioNode,
+             HTMLZipFile: HTML5AppNode,
+             DocumentFile: DocumentNode}
+
+# Long-Range TODOs
+# -- package up images as zip files (using build_carousel)
+
+def guess_extension(url):
```
kollivier: I think this function and `guess_type` should be moved into `ricecooker.utils.paths` as that is where our file / directory-related functions are going.

ricecooker/utils/add_file.py
```diff
@@ -0,0 +1,204 @@
+import requests
```
kollivier: Maybe for the module name, we should do `nodes` rather than `add_file`, so we can put any additional functions that deal with nodes in here?

dragondave: Sounds good to me.

ricecooker/utils/carousel.py
```diff
@@ -0,0 +1,174 @@
...
+from bs4.element import Tag
+from urllib.parse import urlsplit
+from itertools import zip_longest
+import requests
+import hashlib
+import shutil
+import os
+from ricecooker.utils import add_file
+
+html = """
```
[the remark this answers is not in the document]
dragondave: What's the best way of referring to this HTML file's location from inside code? I can't work out if/where pip is installing it to, I can't see it in my site-packages.

ricecooker/utils/add_file.py
```diff
@@ -0,0 +1,204 @@
...
+    "Placeholder filetype for audio that requires transcoding"
+
+DOWNLOAD_FOLDER = "__downloads"  # this generates completed files
+metadata = {}
+have_setup = False
+
+def setup_directory():
+    global have_setup
+    if have_setup: return
+    have_setup = True
```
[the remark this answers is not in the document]
dragondave: Good question. Replaced with try...except.

ricecooker/utils/carousel.py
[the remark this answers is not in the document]
dragondave: Ivan's taken me through how to do this, but since there's work on a new solution for image libraries, I think it's not worth adding this to the codebase in hindsight. Removed carousel.py.

ricecooker/utils/add_file.py
dragondave: Moved guess_extension to paths.get_extension.

For guess_type, I disagree that paths is a good place -- it only very incidently cares about filenames, mostly caring about mimetypes and file contents. I've left it as-is.

ricecooker/utils/add_file.py
```diff
@@ -0,0 +1,204 @@
...
+        return ""
+    ext = "." + filename.split(".")[-1].lower()
+    if "/" in ext:  # dot isn't in last part of path
+        return ""
+    return ext
+
+def create_filename(url):
+    return hashlib.sha1(url.encode('utf-8')).hexdigest() + guess_extension(url)
+
+def download_file(url):
```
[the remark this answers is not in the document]
dragondave: Critically, for this specific case, downloader would lose information about the mimetype of the file we've just downloaded, which is important to correctly inferring what sort of file the document is.

ricecooker/utils/add_file.py
dragondave: I don't know enough about the internals of the ricecooker process to reimplement this -- there were a number of issues such as requirements that files needed to end with particular extensions.

We'd need to ensure that the conversion happened exactly once, even if we're not routinely converting all the videos that are already in the correct format.

As it stands it's a small amount of complexity that I feel could be ignored.

dragondave:
I've addressed the comments -- thanks for the review!

I think there's a good overarching question of whether ricecooker.utils is the correct place for them. I initially thought that it probably wasn't, as they're useful to cheffers rather than being core parts of the ricecooker project, but after discussion with Ivan it was suggested I put them here.

If you think they don't particularly sit comfortably here, then I'm happy for them to live in another repository that can be imported by individual chefs with a quick `pip install`.

kollivier:
I think the main thing is that if we want this in ricecooker, we probably want to work more on making the code generalizable to all chefs. `guess_type` and `guess_extension` are good candidates as-is, but the functions in `add_file` probably need some more work. e.g. `download_file` has a benefit over `read` in that it gives us more information, but it has downsides in that it doesn't take advantage of the caching and retry functionality in the ricecooker function, which are important if we are to recommend this for general usage. 

One thing we could do is to modify `read` to add an `return_headers=False`, and when True, have it return a tuple of headers and content instead of just the content. Then `download_file` could get the content type from the headers, then save the returned content to the desired filename. Then `download_file` now benefits from the retry and caching, which I think would probably be helpful to your own chefs as well.

I think it is good that we're looking at things from that lens and working through these issues, though. I think ricecooker and cheffing are reaching a point where we want to start considering opening these things up to a broader audience, and that will require taking a close look at how we do some things and deciding on things like best practices and standardized approaches to common tasks.

PR #206, opened 2019-04-08
Sphinx Compilation

dragondave:
Working on ReadTheDocs -- https://ricecooker.readthedocs.io/en/oldsphinx/

Removes nbsphinx requirement (by migrating Jupyter notebooks to RST), uses modern way of importing recommonmark. 

Has complete but fairly manual python API docs -- update `ricecooker.rst` if there's a new file. It'd be nice if this was fixed in the future.

Some docstrings modified to suppress errors and improve rendering, but no major attempt ma

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,5 @@ tests/testcontent/
 
 # Pycharm project settings
 .idea/
+
+cache.sqlite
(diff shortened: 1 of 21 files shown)


kollivier:
@ivanistheone What do you think of this? I think it would be good to start getting our docs on RTD! I'm not familiar with how this is setup, so I'm prefer not to merge without review from you on this.

ivanistheone:
This is amazing work @dragondave  !

This needed to be done for the longest time, and you figued out a way to combine md + rst in a civilized manner, so much better than the hacks I had in place before.

PR #208, opened 2019-04-12
Slideshow content kind

nucleogenesis:


ricecooker/classes/nodes.py
```diff
@@ -798,3 +798,56 @@ def truncate_fields(self):
...
+            description (str): description of content (optional)
+            files ([<SlideImageFile>]): images associated with slides
+            thumbnail (str): local path or url to thumbnail image (optional)
+            extra_fields (dict): any additional data needed for node (optional)
+            domain_ns (str): who is providing the content (e.g. learningequality.org) (optional)
+
+    """
+    kind = content_kinds.SLIDESHOW
+
+    def __init__(self, source_id, title, slideshow_data=None, **kwargs):
```
ivanistheone: need `license` as positional arg here after `title`

ricecooker/classes/nodes.py
```diff
@@ -798,3 +798,56 @@ def truncate_fields(self):
...
+            slideshow_data.append(
+                {
+                    'caption': file.caption,
+                    'sort_order': idx,
+                    'checksum': checksum,
+                    'extension': ext
+                }
+            )
+
+        super(SlideshowNode, self).__init__(source_id, title, extra_fields=slideshow_data, **kwargs)
```
ivanistheone: 1. need to pass along `license` after `title` to base class' `__init__`  

2. also note extra_fields should be a dictionary, not a list (see https://github.com/learningequality/ricecooker/blob/master/ricecooker/classes/nodes.py#L395)

ricecooker/classes/nodes.py
```diff
@@ -798,3 +798,56 @@ def truncate_fields(self):
...
+
+    def validate(self):
```
ivanistheone: could also call the base class `validate`

ricecooker/classes/files.py
```diff
@@ -371,6 +371,18 @@ def process_file(self):
...
+    allowed_formats = [file_formats.JPG, file_formats.JPEG, file_formats.PNG]
+    is_primary = True
+
+    def __init__(self, path, caption="", **kwargs):
+        self.caption = caption
+        self.path = path.strip()
+        super(DownloadFile, self).__init__(**kwargs)
+
+    def get_preset(self):
+        return format_presets.SLIDE_IMAGE
```
ivanistheone: SLIDE_IMAGE --> SLIDESHOW_IMAGE

tox.ini
```diff
@@ -16,6 +16,7 @@ setenv =
     PYTHONPATH = {toxinidir}
 commands =
     pip install -U pip
+    pip install -U git+https://github.com/nucleogenesis/le-utils.git@slideshow-content-type#egg=le-utils
```
ivanistheone: @nucleogenesis  You don't need this anymore --- so can remove when you next work on this PR

nucleogenesis:
@ivanistheone sorry to return to this a while after we discussed it - but in our last discussion I believe that we got all of the issues resolved together. Is this ready to merge?

ricecooker/classes/nodes.py
```diff
@@ -798,3 +798,84 @@ def truncate_fields(self):
...
+    def add_file(self, file_to_add):
+        """
+        Add a the SlideImageFile to the node's files and append slideshow metadata
+        to extra_fields['slideshow_data'] (list).
+        Args: file (SlideshowNode or ThumbnailFile): file model to add to node
+        Returns: None
+        """
+        from .files import File, SlideImageFile
+        assert isinstance(file_to_add, File), "Files being added must be instances of a subclass of File class"
+        # TODO asserr SlideshowNode OR ThumbnailFile
```
ivanistheone: This would be good to add here to make sure only `SlideshowFile` and `ThumbnailFile` can be added to `SlideshowNode`s

ricecooker/classes/nodes.py
```diff
@@ -798,3 +798,84 @@ def truncate_fields(self):
...
+            file_to_add.node = self
+            self.files.append(file_to_add)
+
+    def validate(self):
+        print('in SlideshowNode validate')
+        from .files import SlideImageFile
+        try:
+            assert any(self.files), "Assumption Failed: Slideshow does not have any slideshow image files."
+            assert all(filter(lambda f: isinstance(f, SlideImageFile), self.files)), "Assumption Failed: Slideshow files must all be of type SlideImageFile."
+            # TODO: could also be ThumbnailFile    ^
```
ivanistheone: Validation will fail if you add a `ThumbnailFile` to the node now --- so need to change line 877 to be an OR check

tests/test_data.py
```diff
@@ -148,3 +148,20 @@ def test_exercise_to_dict(exercise, exercise_data):
         assert key in exercise_dict, "Key {} is not found in to_dict method".format(key)
     for key, value in exercise_dict.items():
         assert value == exercise_data.get(key), "Mismatched {}: {} != {}".format(key, value, exercise_data[key])
+
+def test_slideshow_to_dict(slideshow, slideshow_data):
+    slideshow_dict = slideshow.to_dict()
+    extra_fields = json.loads(slideshow_dict['extra_fields'])
+    assert len(extra_fields['slideshow_data']) == 10, 'wrong num slides'
+    # TODO check extra_fields
```
ivanistheone: It might be useful to add some checks here to make sure the `extra_fields['slideshow_data']` dict has appropriate fields --- explicitly check the data we send is what Studio expects.

ricecooker/classes/nodes.py
nucleogenesis: Hey @ivanistheone 

I'm curious if I'm misunderstanding the expected functionality of that lambda line at 877. My understanding is that it should parse all of the `self.files` passing each as `f` to the lambda, which then evaluates the `isinstance(...` expression for each file `f`.

However:

```
assert all(filter(lambda f: isinstance(f, SlideImageFile), self.files)), "Assumption Failed: Slideshow files must all be of type SlideImageFile."
```

Is not working functionally the same as:

```
for file in self.files:
                assert isinstance(file, SlideImageFile), "Assumption Failed: Slideshow files must all be of type SlideImageFile."
```

 Because - when I went into the tests and added ThumbnailFiles to the test Slideshows, there was no failure at all - which meant that the first assert that used the filter is not properly asserting against all files.

The second `for x in y` type example, however, fails as it should when it gets to the ThumbnailFile.

I also encountered this exact same issue elsewhere while writing tests and I'll make the note there.

tests/test_data.py
nucleogenesis: @ivanistheone More on the ineffectiveness of the `assert(filter...` 

I tried checking all of the slideshow_data items like so (written in place of your TODO highlighted above):

```
expected_field_keys = { 'caption', 'descriptive_text', 'checksum', 'sort_order', 'extension'}
assert all(filter(lambda f: set(f.keys()) == expected_field_keys), extra_fields['slideshow_data'])), "error message"
```

But when I set the second line to NOT EQUAL like:

```
expected_field_keys = { 'caption', 'descriptive_text', 'checksum', 'sort_order', 'extension'}
assert all(filter(lambda f: set(f.keys()) != expected_field_keys), extra_fields['slideshow_data'])), "error message"
```

The tests still passed.

However, this:

```
expected_field_keys = { 'caption', 'descriptive_text', 'checksum', 'sort_order', 'extension'}
    for slide_data in extra_fields['slideshow_data']:
        assert set(slide_data.keys()) == expected_field_keys, 'extra_fields is missing expected fields'
```

Fails as expected when I switch the logic.

Am I not understanding the `assert all(filter())` code somehow?

ricecooker/classes/nodes.py
ivanistheone: I think the confusion is with the `filter` function which is not needed for this check. Instead what we'd want is something more like: 
```
assert all([isinstance(f, SlideImageFile) or isinstance(f, ThumbnailFile) for f in self.files]), \
               "SlideshowNode files must be of type SlideImageFile or ThumbnailFile."
```
but [your version with the for loop](https://github.com/nucleogenesis/ricecooker/blob/slideshow-content-kind/ricecooker/classes/nodes.py#L881-L882) is just as good—no need to try fit everything in on one line...

tests/test_data.py
ivanistheone: > Am I not understanding the assert all(filter()) code somehow?

Well it's not so much a misunderstanding as it's not necessary --- I think I may have copy-pasted the code from another place filter was useful but here you don't need.

The check here check could be
```
assert all([set(sd.keys()) == expected_field_keys for sd in extra_fields['slideshow_data']]), 'extra_fields is missing expected fields'
```
which is exactly the same as the for loop:
```
    for slide_data in extra_fields['slideshow_data']:
        assert set(slide_data.keys()) == expected_field_keys, 'extra_fields is missing expected fields'
```

ricecooker/classes/nodes.py
```diff
@@ -798,3 +798,89 @@ def truncate_fields(self):
...
+
+            #
+            # Add node->file link
+            file_to_add.node = self
+            self.files.append(file_to_add)
+
+    def validate(self):
+        from .files import SlideImageFile, ThumbnailFile
+        try:
+            assert any(self.files), "Assumption Failed: Slideshow does not have any slideshow image files."
```
ivanistheone: I think you want this to be something like
```
assert [f for f in self.files if isinstance(f, SlideImageFile)], \
    "Assumption Failed: SlideshowNode must have at least one SlideImageFile file."
```

here we're using the fact that empty list `[]` is falsy and non-empty list is truty

ivanistheone:
@nucleogenesis  After looking at your comments, I realized there is something really weird about `filter` outputs --- the output of a filter call is a "filter object" which is list-like, but not quite, since filter objects don't behave like lists with respect to truthyness:

```
>>> mylist = ['foo', 'bar', 'baz']

>>> bool(filter(lambda s: s=='nonexitantkey', mylist))
True  # User expected False as in

>>> bool(list(filter(lambda s: s=='nonexitantkey', mylist)))
False
```

So I suspect the original author of the validation logic added the extra `any` and `all` to avoid this gotcha. 
Here is a little PR on top of your PR to reduce the use of all(filter pattern throught the code base:
https://github.com/nucleogenesis/ricecooker/pull/2

ricecooker/classes/nodes.py
nucleogenesis: The lines below that take care of that logic, right? I also included the assertion that it is SlideImageFile OR ThumbnailFile in the lines below using a for loop.

nucleogenesis:
> @nucleogenesis After looking at your comments, I realized there is something really weird about `filter` outputs --- the output of a filter call is a "filter object" which is list-like, but not quite, since filter objects don't behave like lists with respect to truthyness:
> 
> ```
> >>> mylist = ['foo', 'bar', 'baz']
> 
> >>> bool(filter(lambda s: s=='nonexitantkey', mylist))
> True  # User expected False as in
> 
> >>> bool(list(filter(lambda s: s=='nonexitantkey', mylist)))
> False
> ```
> 
> So I suspect the original author of the validation logic added the extra `any` and `all` to avoid this gotcha.
> Here is a little PR on top of your PR to reduce the use of all(filter pattern throught the code base:
> [nucleogenesis#2](https://github.com/nucleogenesis/ricecooker/pull/2)

Awesome @ivanistheone thank you! I merged that PR and applied the changes requested above.

That is weird that filter() returns its own type rather than a list - just not what I'd expect it to do. 

In any case - is there anything remaining that ought to be changed or added?

PR #209, opened 2019-05-08
Helper script for quick debugging of HTML5Apps using local Kolibri

ivanistheone:
All credit goes to @dragondave for coming up with the idea!

### DIFF
--- a/ricecooker/utils/downloader.py
+++ b/ricecooker/utils/downloader.py
@@ -193,13 +193,7 @@ def download_static_assets(doc, destination, base_url,
     def js_content_middleware(content, url, **kwargs):
         if js_middleware:
             content = js_middleware(content, url, **kwargs)
-
-        # Polyfill localStorage and document.cookie as iframes can't access
-        # them
-        return (content
-            .replace("localStorage", "_localStorage")
-            .replace('document.cookie.split', '"".split')
-            .replace('document.cookie', 'window._document_cookie'))
+        return content
 
     def css_node_filter(node):
         return "stylesheet" in node["rel"]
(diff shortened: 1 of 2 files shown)


ricecooker/utils/downloader.py
```diff
@@ -193,13 +193,7 @@ def download_assets(selector, attr, url_middleware=None,
     def js_content_middleware(content, url, **kwargs):
         if js_middleware:
             content = js_middleware(content, url, **kwargs)
-
-        # Polyfill localStorage and document.cookie as iframes can't access
-        # them
-        return (content
-            .replace("localStorage", "_localStorage")
-            .replace('document.cookie.split', '"".split')
-            .replace('document.cookie', 'window._document_cookie'))
```
kollivier: Can we separate this change out? I think we need to figure out how to handle the `document.cookie.split` change in a way that doesn't break old Kolibris.

ricecooker/utils/kolibripreview.py
```diff
@@ -0,0 +1,40 @@
...
+    indexpath = os.path.join(srcdir, 'index.html')
+    if not os.path.exists(indexpath):
+        print('Missing index.html file in', srcdir)
+        return False
+    return True
+
+
+def main(args):
+    """
+    Command line utility for previewing HTML5App content in Kolbri.
```
kollivier: Can we add a note about the fact that this works by replacing an existing HTML5 zip in the local Kolibri installation? I know the destzip arg docstring alludes to it, but I think it would be good to make it clear in the tool documentation.

PR #213, opened 2019-06-28
Integrate autothumbs functions in classes/files.py

ivanistheone:
Document functionality of each function without changing their signature or operation.

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -873,7 +873,7 @@ class ExtractedThumbnailFile(ThumbnailFile):
             copy_file_to_storage(filename, tempf.name)
             os.unlink(tempf.name)
             FILECACHE.set(key, bytes(filename, "utf-8"))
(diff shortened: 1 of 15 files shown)


ricecooker/classes/files.py
```diff
@@ -183,7 +201,13 @@ def get_hash(filepath):
     return file_hash.hexdigest()
 
 
+
+
 def compress_video_file(filename, ffmpeg_settings):
+    """
+    Calls the pressurecooker funciton `compress_video` to compress filename (source)
```
kollivier: Typo in function.

kollivier:
After merging in the pressurecooker thumbnail changes and running this PR locally, I'm seeing a couple errors, similar to the one Travis is showing with 'invalid file' messages. I get that for PDF, mp3, and zip. Pressurecooker tests are passing. Not sure if it's an issue with the files themselves or how they are being processed, but would you have some time to look into this @ivanistheone?

PR #214, opened 2019-07-09
HTML writer: Don't read from the url if it has already been imported

jayoshih:
Since `write_contents` does a check before writing the content to the zipfile, it seems like it would be faster to have `write_url` do a similar check before reading the url


ivanistheone:
nice; thx for fix

PR #219, opened 2019-08-13
Export channel tree as JSON

kollivier:
This can help with local review and testing during chef development, so that you can check the generated tree without having to upload to Studio first. This will also help us as we build out JSON tree diffing tools, as this can allow us to do local diffs if we have a JSON tree file to compare against.


kollivier:
Not ready to land yet as it needs tests, but wanted to get some eyes on it and discuss general usefulness.

benjaoming:
:+1: 

Being new at this, I found it nice to have previous milestone metadata results stored as JSON in a repo. It made it easy to see the real impact of changes. I guess this is only relevant for relatively small sets of data.

It partly overlaps with the `JsonTreeChef`, making it superfluous in some cases: The code of the SHLS chef creates JSON and then imports it -- but seemingly the JSON could just be a bi-product intended to keep record of changes, and the chef could create a Channel object directly, exporting JSON with `channel_as_json` instead for record-keeping. See: https://github.com/learningequality/sushi-chef-shls/blob/03f2fa9d8d500dbb7c2f8c1aa89d41651ca4bf06/sushichef.py#L581

For introspecting generated data, perhaps implementing `__repr__` methods on all Channel tree/node related structures would be sufficient - so running `print(channel)` would produce human readable contents?

On a more general level, this could perhaps be implemented through the idea of (de-)serialization: https://docs.djangoproject.com/en/3.0/topics/serialization/

ivanistheone:
@benjaoming There has been some work on this since this PR. We've decided to make the "studio json tree" export part of standard ricecooker workflow in order to keep a local log of the entire tree that was uploaded to studio on each run. Here is a [POC implementation](https://github.com/ivanistheone/ricecooker/commit/8f66b0a814d4fe45328503de902efff230977fcc#diff-c7f133f8e73bff48aa135316e3e60d3cR201) of how this could work.

We're keeping track of next steps on this as part of [this project](https://www.notion.so/learningequality/Ricecooker-optimizations-e6713176dc894cd882e62cb26f060d04). Basically having ricecooker output an entire studio json tree as part of the run will allow us to upload the whole channel metadata in one show instead of doing folder by folder as we currently do (which takes many hours for large chefs and is a known limitation).


> It partly overlaps with the JsonTreeChef, making it superfluous in some cases
> 

You're right about the overlap with `ricecooker_json_tree` files, but the use case is slightly different. The purpose of `ricecooker_json_tree` is to store the input data for ricecooker, while the `studio_json_tree` is to store the output data (=input data for studio).

Both could be useful for debugging, but by standardizing on the `studio_json_tree` that is auto-generated each time the chef runs we'll be sure to keep "tree logs" for all cases.


-----

@kollivier I'm going to close this PR since it's superseded by something like the [POC implementation](https://github.com/ivanistheone/ricecooker/commit/8f66b0a814d4fe45328503de902efff230977fcc#diff-c7f133f8e73bff48aa135316e3e60d3cR201).

PR #224, opened 2019-09-17
Updates to json workflow

ivanistheone:
- Handle tags instead of dropping them
  - Add support for Slideshow content kind

### DIFF
--- a/ricecooker/utils/jsontrees.py
+++ b/ricecooker/utils/jsontrees.py
@@ -21,6 +21,7 @@ AUDIO_NODE = content_kinds.AUDIO
 EXERCISE_NODE = content_kinds.EXERCISE
 DOCUMENT_NODE = content_kinds.DOCUMENT
 HTML5_NODE = content_kinds.HTML5
+SLIDESHOW_NODE = content_kinds.SLIDESHOW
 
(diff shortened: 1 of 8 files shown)


ricecooker/utils/jsontrees.py
```diff
@@ -205,7 +230,7 @@ def build_tree_from_json(parent_node, sourcetree):
 
 def add_files(node, file_list):
     EXPECTED_FILE_TYPES = [VIDEO_FILE, AUDIO_FILE, DOCUMENT_FILE, EPUB_FILE,
-                           HTML5_FILE, THUMBNAIL_FILE, SUBTITLES_FILE]
+                           HTML5_FILE, THUMBNAIL_FILE, SUBTITLES_FILE, SUBTITLES_FILE]
```
kollivier: Was this supposed to be SLIDESHOW_FILE?

ivanistheone: Thx for spotting that. I was just testing this with some sample code that Jacob sent me, and ran into that error. Fixed now (rebased to merge typo-fix commits into single one).

Slideshow via JSON works and is identical to the original one made via classes:
http://35.229.70.110/en/learn/#/topics/c/5d072a67058751ec9edaee96021d201a

Note: slideshows are officially out in K 0.13

PR #230, opened 2019-10-29
Report API Error message

dragondave:
Minor tweak to output error messages (e.g. "user is not authorised") when accessing channel REST api.


dragondave:
(Possibly should use response.text to make it not a byte string, for prettiness' sake)

ivanistheone:
combined into https://github.com/learningequality/ricecooker/pull/229

PR #234, opened 2019-12-17
H5P as separate content kind

ivanistheone:
Added a new `H5PAppNode` content kind for `.h5p` files.


jayoshih:
Just need to update the version, so I'll let you merge once that's done :)

PR #242, opened 2020-01-30
Make --stage behavior the default, and add a --deploy flag…

kollivier:
…to make overwriting tree contents an explicit option.

### DIFF
--- a/ricecooker/chefs.py
+++ b/ricecooker/chefs.py
@@ -95,6 +95,9 @@ class BaseChef(object):
         args_namespace, options_list = self.arg_parser.parse_known_args()
         args = args_namespace.__dict__
 
+        if args['stage_deprecated']:
+            config.LOGGER.warning('DEPRECATION WARNING: --stage is now default, so the --stage flag has been deprecated and will be removed in ricecooker 1.0.')
+
         # Make sure token is provided. There are four possible ways to specify:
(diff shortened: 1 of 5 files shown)


ricecooker/chefs.py
```diff
@@ -290,16 +293,15 @@ def run(self, args, options):
 
     def main(self):
         args, options = self.parse_args_and_options()
+        if args['stage_deprecated']:
+            config.LOGGER.warning('DEPRECATION WARNING: --stage is now default, so the --stage flag has been deprecated and will be removed in ricecooker 1.0.')
```
ivanistheone: @kollivier Could you move this warning to here:
https://github.com/learningequality/ricecooker/blob/master/ricecooker/chefs.py#L94

this way all the subclasses (SushiChef, JsonTreeChef, etc) will behave the same way.

ivanistheone:
Regarding the failing tests, it's because the current ricecooker CLI interface tries to match the old doc-opt CLI interface:
https://github.com/learningequality/ricecooker/blob/master/tests/test_argparse.py#L17-L52



I think by now we can totally get rid of this old docopt-compatibility mode now...  it's been years since last chef used the `python -m ricecooker chefscript.py` way of running scripts.

I can open a folloup PR to get rid of those doc-opt compatibility tests, so feel free to ignore.

PR #247, opened 2020-02-18
Create a file log of run in logs folder and ensure all loggers…

kollivier:
…and handlers use the same logging level.

### DIFF
--- a/ricecooker/chefs.py
+++ b/ricecooker/chefs.py
@@ -1,15 +1,18 @@
 import argparse
+from datetime import datetime
+import logging
 import os
 import sys
 from importlib.machinery import SourceFileLoader
 
 
+
 from . import config
 from .classes.nodes import ChannelNode
-from .commands import uploadchannel, uploadchannel_wrapper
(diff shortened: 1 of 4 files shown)


ivanistheone:
This PR consolidates all ricecooker logging functionnality into a single place -- the `config_logger` method on the chef class.

There are three loggers:
  - standard stderr logging (what we always had)
  - a new log to file feature (turned on by default, same log level but includes timstamps)
  - sushibar remote logging (same as before, but with less aggressive error messages when we fail to connect --- these are non-critical errors so we just need to let users know the remote logging feature is disabled). This could be a first step towards reducing/removing need for sushibar.

Implementation note:  if a chef script uses `ricecooker.config.LOGGER` before the chef has started running, the logs will be handled by a default temporary stream logger that outptus to stderr (with log level info). This temporary logger is removed once chef runs.

ivanistheone:
@kollivier The github user interface doesn't allow me to request you as reviewer since you opened the PR, so plz reply as comments when you've had time to look at this one.

kollivier:
LGTM!

PR #248, opened 2020-02-28
Implement --sample command line argument

ivanistheone:
This is option can be used to "test" a large channel by uploading only a sample of size `SIZE` content nodes from the tree.

Note: the channel's source id is modified so content is uploaded to a different `channel_id`. This is intentional used to avoid any chance of overwriting an existing production channel.

Merge only after logging updates https://github.com/learningequality/ricecooker/pull/247


kollivier:
@ivanistheone I think you may need to merge this PR with master after I merged #247 and then push again, as it still is showing the logging changes.

PR #249, opened 2020-03-02
Add dry-run flag to only run channel creation code, and make sushibar…

kollivier:
… reporting off by default.


kollivier:
My assumption is that --nomonitor is infrequently used and not used in automated scripts as I don't recall seeing it used, so I did not leave the old flag for compatibility. I can add that if needed.

kollivier:
BTW, I added this flag because it's often quite a few runs during initial development before I get to the point of being ready to upload the channel. During this phase, I didn't want it creating a channel on the server just to do a run of the chef.

PR #257, opened 2020-04-05
Add some more explanations and guidance for those new to Kolibri and …

kollivier:
…its concepts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,27 @@
 ricecooker
 ==========
-The `ricecooker` library is a framework for creating Kolibri content channels and
-uploading them to [Kolibri Studio](https://studio.learningequality.org/), which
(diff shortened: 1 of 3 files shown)


kollivier:
Note that I think there is some overlap between the docs already in the `README.md` file and content in `docs/platform/introduction.md`, so I think we can probably do some de-duplication there. Also, are things like the CSV workflow fully supported, i.e. will we be okay if people read these docs and start trying to use it? If not, we may want to make it less visible in the docs. 

Also wondering if it makes sense to link / reference or update some parts using Danielle's content integration docs?

ivanistheone:
@kollivier Can you please rebase to current master. I want to see if test error is genuine or just because outdated.

PR #262, opened 2020-04-27
Bug fix and improvement for log handling

### LINKED ISSUE #259: Empty void_error_log.log file appears after chef run
* ricecooker version: 0.6.42
* Python version: 3.5
* Operating System: linux on vader

### Description

Running the KA chef from the currculumKey branch and I see an empty `void_error_log.log` created in root of chef repo.

LOGGER is imported using the "convenience import" method that is used in most chefs:
https://github.com/learningequality/sushi-chef-khan-academy/blob/curriculumKey/json_chef.py#L11

### What I Did

Ran the chef with -v flag (default).

### Expected

Log to be created in logs/ dir + logging to stdout. No separate file for errors.

### Actual

Empty `void_error_log.log` file.

### Cause?

Seems to be a side effect of the optional separate file for errors, which should have been off by default: https://github.com/learningequality/ricecooker/blob/master/ricecooker/config.py#L74
but it seems file still gets created.

benjaoming:
* Fixes #259
* Gathers some more of the logging behavior in `setup_logging`
* Updates config `logs/<date>.log` to contain all configured logging (previously only `config.LOGGERS`).
* Adds new `logs/<date>.err.log` with only errors

There is an alternative version here that also lints `chefs.py` and `config.py` because they failed flake8: https://github.com/learningequality/ricecooker/compare/master...benjaoming:logger-fix

### DIFF
--- a/ricecooker/chefs.py
+++ b/ricecooker/chefs.py
@@ -82,6 +82,7 @@ class BaseChef(object):
                             help='Immediately deploy changes to channel\'s main tree. This operation will delete the previous channel content once upload completes. Default (recommended) behavior is to post new tree for review.')
         parser.add_argument('--publish', action='store_true',         help='Publish newly uploaded version of the channel')
         parser.add_argument('--sample', type=int, metavar='SIZE',     help='Upload a sample of SIZE content nodes from the channel')
+
         # [OPTIONS] --- extra key=value options are supported, but do not appear in help
         self.arg_parser = parser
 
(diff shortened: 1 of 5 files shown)


ivanistheone:
@benjaoming I changed the default level to logging.WARNING for the .err.log file. The filename will be a little confusing, but I think this will make this file more useful --- basically this is the things that chef author MUST look at after the chef runs.

benjaoming:
@ivanistheone yes I absolutely agree on that change, thanks for that!

I guess it's okay that the file remains to be called `.err` -- strictly speaking, it would also have the `CRITICAL` log level :)

PR #264, opened 2020-04-28
Check for commonly used template domain and IDs

kollivier:
While going through the docs I realized we have been pretty consistent about the template placeholders we have used in the docs and examples, so I thought an easy fix is just to check for them and report them as an error. (Because they've already been used on Studio they won't work anyway.)

### DIFF
--- a/ricecooker/chefs.py
+++ b/ricecooker/chefs.py
@@ -260,6 +260,20 @@ class BaseChef(object):
             return channel
 
         elif hasattr(self, 'channel_info'):
+            # Before going any further, make sure there aren't template id values in their channel info.
+            template_domains = ['<yourdomain.org>']
+            using_template_domain = self.channel_info['CHANNEL_SOURCE_DOMAIN'] in template_domains
+            if using_template_domain:
+                config.LOGGER.error("Template source domain detected. Please set CHANNEL_SOURCE_DOMAIN before running this chef.")
+
+            template_ids = ['<unique id for the channel>', '<yourid>']
+            using_template_source_id = self.channel_info['CHANNEL_SOURCE_ID'] in template_ids
+            if using_template_source_id:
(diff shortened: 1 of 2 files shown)


ivanistheone:
Nice idea. We should standardize all examples/ and sample code to use these values.

One issue is this won't catch chefs that implement a custom `get_channel` method like these examples https://ricecooker.readthedocs.io/en/latest/examples/languages.html although I think this means we should rewrite these examples to use the "standard" approach with `channel_info` attribute and template-standard values '<yourdomain.org>' and  '<unique id for the channel>' 

I guess by the time ricecooker users are creating chefs that implement custom `get_channel` or use LineCook, JsonTreeChef, they probably know how to set source domain and channel source id ;)

ricecooker/chefs.py
```diff
@@ -260,6 +260,20 @@ def get_channel(self, **kwargs):
...
+            # Before going any further, make sure there aren't template id values in their channel info.
+            template_domains = ['<yourdomain.org>']
+            using_template_domain = self.channel_info['CHANNEL_SOURCE_DOMAIN'] in template_domains
+            if using_template_domain:
+                config.LOGGER.error("Template source domain detected. Please set CHANNEL_SOURCE_DOMAIN before running this chef.")
+
+            template_ids = ['<unique id for the channel>', '<yourid>']
+            using_template_source_id = self.channel_info['CHANNEL_SOURCE_ID'] in template_ids
+            if using_template_source_id:
+                config.LOGGER.error("Template source ID deteected. Please set CHANNEL_SOURCE_ID before running this chef.")
```
ivanistheone, line 272: I was confused by the word "Template" on first reading wondering what template. Not sure if I have better alternative though...  Example maybe? Default values?

Also "set" is a bit confusing. Maybe "Change from the default". 

Leaving off perchance inspiration comes before tomorrow, but since none of my suggestions are any better than current options, I'm going to merge as-is tmrw AM.

Issue #272, opened 2020-05-16
Python 3.5 dict key ordering breaks cache


ivanistheone:
* ricecooker version: 0.6.43
* Python version: 3.5.3
* Operating System: Linux

### Description

The cache key is determined based on the `download_settings` which is a Python dict.

In Python dict the dict `str({"a":3, "b":4})` could end up in `{"a":3, "b":4}` or `{"b":4, "a":3}`.

The code already tries to be deterministic:
https://github.com/learningequality/ricecooker/blob/cd559e091bc04f2480337ed89ccfe674b43eef70/ricecooker/classes/files.py#L71
but if download settings has deep attibutes, such as 
```
 download_settings = {
...                         'postprocessors': [
...                             {
...                                 'key': 'ExecAfterDownload',
...                                 'exec_cmd': 'ffmpeg -hide_banner -loglevel panic -i {} -b:a 32k -ac 1 {}_tmp.mp4 && mv {}_tmp.mp4 {}',
...                             }
...                         ]
...                     }
```
it that stratrgy fails so we need deeper sorted.

### Solution

https://github.com/ivanistheone/ricecooker/commit/db5f26c1bd3b89272c39ee88bca688f7ecc06461
via https://stackoverflow.com/questions/5884066/hashing-a-dictionary/22003440#22003440

ivanistheone:
Fixed in https://github.com/learningequality/ricecooker/commit/b03357242de47a29f19683dfc12a36be7908b92e#diff-265dcdb28f58f179728e3e42490456a4R73-R75 .

PR #273, opened 2020-05-22
Changes to --update logic to fix #269 and fix for #272

### LINKED ISSUE #269: Known limitations of the --update CLI argument
* ricecooker version: 0.6.43
* Python version: all
* Operating System: all

### Description of problems

Three mostly-unrelated issues related to the use of `--update` flag:

1. It doesn't ever make sense to ever update when using [download_from_web](https://github.com/learningequality/ricecooker/blob/master/ricecooker/classes/files.py#L229) since video files on youtube and vimeo can't be updated.

2. It is often desired to update *only* the thumbnail files for content nodes (e.g. to re-generate topic 4-up thumbs based on new files added), but not the other files (to avoid big downloads). 

3. It doesn't make sense to cache local paths since it's both confusing (changing local file without modifying path will re-use the old file) and unnecessary (we're not not making things faster since files are already local, so just requires moving data into storage again). Note however that caching the COMPRESSED version of local files is useful — to prevent need for re-processing (video transcoding).

4. Caching tempfile-generated zip files is useuless since they will never be re-used.

### Possible solutions

1. Remove option to update in download_from_web

2. Always update thumbnails by default --- this will lead to some unnecessary downloads for network-thumbnails, and extra re-generation of local thumbnails, but these are minor costs (even for huge channels).

3. Disable update behavior for locally cached files (unless compress used).

4. This needs to be addressed as part of general review of temp files and caching (i.e. handled before --update comes into play).

ivanistheone:
This PR contains several maintenance-type changes and fixes for issues https://github.com/learningequality/ricecooker/issues/269 and https://github.com/learningequality/ricecooker/issues/272

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -260,7 +260,7 @@ def download_from_web(web_url, download_settings, file_format=file_formats.MP4, 
     :return: filename derived from hash of file contents {md5hash(file)}.ext
     """
     key = generate_key("DOWNLOADED", web_url, settings=download_settings)
-    if not config.UPDATE and FILECACHE.get(key):
+    if FILECACHE.get(key):
         return FILECACHE.get(key).decode('utf-8')
 
     # Get hash of web_url to act as temporary storage name
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -5,6 +5,7 @@ from cachecontrol.caches.file_cache import FileCache
 import hashlib
 import os
 from PIL import Image
+import json
 from requests.exceptions import MissingSchema, HTTPError, ConnectionError, InvalidURL, InvalidSchema
 import shutil
 from subprocess import CalledProcessError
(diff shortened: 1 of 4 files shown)


ivanistheone:
In this PR:
 - special hack to add deterministic dict serialization to avoid cache misses in py3.5  (fixes https://github.com/learningequality/ricecooker/issues/272)
 - fixes based of three out of the four sub-issues in https://github.com/learningequality/ricecooker/issues/269
     - 1. youtube_dl downloads are immutable, so always use cache https://github.com/learningequality/ricecooker/pull/273/commits/296708e9b8d34e60e3da475cbccdc57c2df1d755
     - 2. thumbnails are now always re-generated, see https://github.com/learningequality/ricecooker/pull/273/commits/6d2d04c622f5a80909a93f85e3af75947945eed2 and https://github.com/learningequality/ricecooker/pull/273/commits/51548a9b59641a9fc53081f1d6bf66fcb42fe732
     - 3. Added [is_valid_url](https://github.com/learningequality/ricecooker/pull/273/commits/1196c288a5e7d4ea2e0d71cb32c2e87b3997cc97#diff-265dcdb28f58f179728e3e42490456a4R146-R155) helper function to distinguish URL paths from local paths, and used it to disable file cache for local paths https://github.com/learningequality/ricecooker/pull/273/commits/bce9f6c2e968468866617a91884041f00f92ca7e


There is also an unrelated change of changing default to `high_resolution=False` so we get 480 videos by default, https://github.com/learningequality/ricecooker/pull/273/commits/ab829fa1e4f01bbd64724bea2cc615a04ce8141b

ivanistheone:
The `download_settings` key `postprocessors` is special since it is a list of dicts and requires special serialization. Upon inspection no other keys are dict-like, but need to watch out for this — in case other nested dicts exist, serialization will not be detemrinistic and caching will fail.

PR #274, opened 2020-06-09
Encode titles so logging doesn't throw error

jayoshih:
When I run chefs in other languages, it sometimes throws an error on the logging statement, which then prevents those nodes from getting scraped


kollivier:
Could you give more details on the problem? We can output to multiple sources, with different encodings, so we should only encode text when we're about to output to a target requiring byte strings instead of unicode strings. (e.g. the terminal, though Python should be handling conversion for us in that case)

PR #275, opened 2020-06-17
Add support for specifying a Channel tagline.

kollivier:


ricecooker/classes/nodes.py
```diff
@@ -254,6 +254,8 @@ def __init__(self, source_id, source_domain, *args, **kwargs):
         # Map parameters to model variables
         self.source_domain = source_domain
         self.source_id = source_id
+        # TODO: Add this to channel_info? Currently you need to set this after calling get_channel.
+        self.tagline = None
```
jayoshih, line 258: I agree, that this makes sense as part of channel_info. Otherwise, you might need to have implicit knowledge of this being there

PR #276, opened 2020-06-17
Ensure chefs always clean up after themselves.

kollivier:
This ensures any and all tempfiles that get created will be deleted either by the end of the run, or if that failed for some reason, at the beginning of the next run.

### DIFF
--- a/ricecooker/chefs.py
+++ b/ricecooker/chefs.py
@@ -1,7 +1,10 @@
 import argparse
+import atexit
 import logging
 import os
+import shutil
 import sys
+import tempfile
 from datetime import datetime
 from importlib.machinery import SourceFileLoader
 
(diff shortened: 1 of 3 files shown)


ricecooker/chefs.py
```diff
@@ -31,6 +34,28 @@
...
+# from the old run are deleted so that they do not accumulate.
+delete_temp_dir()
+
+# If tempdir is set already, that means the user has explicitly chosen a location for temp storage
+if not tempfile.tempdir:
+    os.makedirs(chef_temp_dir)
+    config.LOGGER.info("Setting chef temp dir to {}".format(chef_temp_dir))
+    # Store all chef temp files in one dir to avoid issues with temp or even primary storage filling up
+    # because of failure by the chef to clean up temp files manually.
+    tempfile.tempdir = chef_temp_dir
```
ivanistheone, line 56: This seems to have the same effect as setting the ENV variable `TMPDIR` (all temp files are created as children of it).

ivanistheone:
Tested using the wikipedia example and works great!

Tempfiles created in `.ricecooker-temp/` during chef run:
Screen Shot 2020-06-17 at 7 48 35 PM

then the `.ricecooker-temp/` is cleaned up on exit:

Screen Shot 2020-06-17 at 7 49 04 PM

PR #277, opened 2020-07-03
Remove dead code around compatibility_mode and unused docopt tests

ivanistheone:
- The `SushiChef` class is now the only one you need (contain combined functionality that was previously split between `BaseChef` and `SushiChef`
 - Re-enabled tests around CLI args and options

### DIFF
--- a/ricecooker/chefs.py
+++ b/ricecooker/chefs.py
@@ -58,7 +58,7 @@ class SushiChef(object):
         else:
             add_parser_help = True
         parser = argparse.ArgumentParser(
-            description="Chef scripts upload content to the Kolibri Studio server.",
+            description="Chef script for uploading content to Kolibri Studio.",
             add_help=add_parser_help,
         )
         self.arg_parser = parser  # save as class attr. for subclasses to extend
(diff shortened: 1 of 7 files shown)


ivanistheone:
NOTE: must merge https://github.com/learningequality/ricecooker/pull/273 first, as this is a followup

ricecooker/chefs.py
```diff
@@ -33,36 +31,39 @@
...
+        subclasses = self.__class__.__mro__[:-2]     # all subclasses after this
+        if any(['__init__' in c.__dict__.keys() for c in subclasses]):
+            add_parser_help = False    # assume subclass' __init__ will add help
+        else:
+            add_parser_help = True
         parser = argparse.ArgumentParser(
-            description="Ricecooker puts your content in the conten server.",
-            add_help=(self.__class__ == BaseChef)  # only add help if not subclassed
+            description="Chef script for uploading content to Kolibri Studio.",
+            add_help=add_parser_help,
```
ivanistheone, line 62: In case anyone is wondering, using `add_help=False` is necessary in cases where a subclass implements its own custom arg parser. This feature is not used much by suchichef classes (I think only MIT blossoms chef and some of Alejando's chefs use it), but used by the  LineCook class (CSV parsing logic).

For passing custom keyword args to chefs, the approach used by @WenyuZhang1992 is much nicer (no need to overload `__init__`):
https://github.com/learningequality/sushi-chef-refugee-response-crisis-advice/blob/master/sushichef.py#L75-L102

PR #290, opened 2020-10-08
Save metadata to csv

richard-dinh:


ricecooker/classes/nodes.py
```diff
@@ -212,6 +216,46 @@ def get_json_tree(self):
...
+ 
+        if not is_channel:
+            # Build out tag string
+            current_level = self.title
+            # prevent list from being passed by reference. Create new instance of list
+            structure_list = list(topic_structure)
+            if len(structure_list) < 1:
+                structure_list = [current_level]
+            else:
+                structure_list.append(current_level)
```
kollivier, line 233: The `if len(structure_list) < 1:` case should no longer be needed, because we're no longer using a reference to a list as the default value for the arg.

ricecooker/commands.py
```diff
@@ -62,7 +66,12 @@ def uploadchannel(chef, command='uploadchannel', update=False, thumbnails=False,
 
     # Get domain to upload to
     config.init_file_mapping_store()
-
+    
+    # Path to chefdata
+    CSV_DIR = os.path.join(Path(__file__).parents[1], 'chefdata', 'data', 'content_metadata.csv')
```
kollivier, line 71: This is a full path to a file, not a directory, so we probably want to name this something other than `CSV_DIR`.

Also, `__file__` is the absolute path to `riececooker/commands.py`, not the root of the chef. This will try to save the file inside the Python module. You can use `os.getcwd()` to get the current directory, or even just leave it as a relative path, like you do in `save_metadata_file_as_csv`.

ricecooker/commands.py
```diff
@@ -89,6 +98,29 @@ def uploadchannel(chef, command='uploadchannel', update=False, thumbnails=False,
     if hasattr(chef, 'download_content'):
         chef.download_content()
 
+    # TODO load csv if exists
+    if os.path.exists(CSV_DIR):
```
kollivier, line 102: I'd suggest making a `load_metadata_file_from_csv` function as a corollary  to `save_metadata_file_to_csv` and putting it in the chef instead of having the code here.

ricecooker/chefs.py
```diff
@@ -281,10 +282,49 @@ def save_channel_tree_as_json(self, channel):
...
+            return
+            
+        is_channel = isinstance(contentNode, ChannelNode)
+
+        if not is_channel:
+            # Add modifications to contentNode
+            if contentNode.source_id in metadata_dict:
+                contentNode.node_modifications = metadata_dict[contentNode.source_id]
+            for child in contentNode.children:
+                self.save_modifications(child, metadata_dict, metadata_csv)
```
kollivier, line 323: This for loop runs in all cases, so I think you can move it outside of the if / else block and just always run it.

ricecooker/utils/youtube.py
```diff
@@ -47,7 +47,7 @@ def _get_youtube_info(self, use_proxy=True, use_cache=True, options=None):
         youtube_info = None
         # 1. Try to get from cache if allowed:
         if os.path.exists(self.cache_path) and use_cache:
-            LOGGER.info("==> [%s] Retrieving cached information...", self.__str__(), self.cachename)
+            # LOGGER.info("==> [%s] Retrieving cached information...", self.__str__(), self.cachename)
```
kollivier, line 50: Was this getting too noisy? If you switch to using `LOGGER.debug`, it should stop printing this by default.

ricecooker/chefs.py
```diff
@@ -281,10 +282,49 @@ def save_channel_tree_as_json(self, channel):
...
+        ]
+        metadata_csv.writerow(headers)
+
+        channel.save_channel_children_to_csv(metadata_csv)
+
 
     def save_chef_data(self):
         json.dump(self.CHEF_RUN_DATA, open(config.DATA_PATH, 'w'), indent=2)
 
+    def save_modifications(self, contentNode, metadata_dict = {},metadata_csv = None):
```
kollivier, line 311: Could we call this `apply_modifications` instead?

ricecooker/chefs.py
```diff
@@ -281,10 +282,49 @@ def save_channel_tree_as_json(self, channel):
...
 
+    def save_channel_metadata_as_csv(self, channel):
+        # create data folder in chefdata
+        # config.LOGGER.info('SAVE METADATA TO CSV')
+        DATA_DIR = os.path.join('chefdata', 'data')
+        if not os.path.isdir(DATA_DIR):
+            os.makedirs(DATA_DIR)
+        metadata_csv = csv.writer(open(os.path.join(DATA_DIR, 'content_metadata.csv'), 'w', newline=''))
+        headers = [
+            'Source_id',
```
kollivier, line 293: Can we call this "Source ID" to be consistent with the formatting of the other field names?

kollivier: Also, it may be a good idea to create constants for these in `ricecooker.config`, and use those constants here instead of that actual string. That way, if we want to change the strings, we won't have to search the code for the string.

ricecooker/classes/nodes.py
```diff
@@ -212,6 +216,46 @@ def get_json_tree(self):
...
+                tags_string,
+                self.node_modifications.get('New Tags') or '',         # New Tags
+                ''                                                     # Last Modified
+            ]
+            metadata_csv.writerow(record)
+            for child in self.children:
+                child.save_channel_children_to_csv(metadata_csv, structure_list)
+        else:
+            for child in self.children:
+                child.save_channel_children_to_csv(metadata_csv, topic_structure)
```
kollivier, line 256: Once we remove the need to have two list variables, this for loop can be moved out of the if / else and always run.

ricecooker/classes/nodes.py
richard-dinh: removed this

ricecooker/commands.py
richard-dinh: updated

ricecooker/commands.py
richard-dinh: made the function

ricecooker/utils/youtube.py
richard-dinh: will do. It was giving me a logging error every time I ran the test. I'll change it to debug

ricecooker/chefs.py
```diff
@@ -281,10 +283,57 @@ def save_channel_tree_as_json(self, channel):
         self.CHEF_RUN_DATA['tree_archives']['current'] = filename.replace(os.getcwd() + '/', '')
         self.save_chef_data()
 
+    def save_channel_metadata_as_csv(self, channel):
+        # create data folder in chefdata
+        # config.LOGGER.info('SAVE METADATA TO CSV')
+        DATA_DIR = os.path.join('chefdata', 'data')
+        if not os.path.isdir(DATA_DIR):
+            os.makedirs(DATA_DIR)
```
kollivier, line 291: Not a big issue, but Python 3's `makedirs` can be called with `exist_ok=True` now, so that you don't have to check if the dir exists first. `exist_ok` was added in Python 3.2, so it's safe for any of our Python 3 only code to use. (We don't support Python 3 below 3.4 at this point.)

ricecooker/chefs.py
kollivier: I'm still seeing this as `save_modifications` in the latest version.

ricecooker/chefs.py
```diff
@@ -281,10 +283,57 @@ def save_channel_tree_as_json(self, channel):
...
+        metadata_csv = csv.writer(open(os.path.join(DATA_DIR, 'content_metadata.csv'), 'w', newline='', encoding='utf-8'))
+        metadata_csv.writerow(config.CSV_HEADERS)
+
+        channel.save_channel_children_to_csv(metadata_csv)
+
+    def load_channel_metadata_from_csv(self, metadata_csv = None, metadata_dict = {}):
```
kollivier, line 297: `meatadata_dict = {}` will retain modifications between calls. You can use `dict()` instead, much like the `list()` fix. 

Also, are there cases where we want to pass these variables in to the function? If not, we can just make this function take zero args?

ricecooker/chefs.py
```diff
@@ -281,10 +283,57 @@ def save_channel_tree_as_json(self, channel):
...
+                    if line_new_title != '':
+                        metadata_dict[line_source_id]['New Title'] = line_new_title
+                    if line_new_description != '':
+                        metadata_dict[line_source_id]['New Description'] = line_new_description
+                    if line_new_tags != '':
+                        tags_arr = re.split(',| ,', line_new_tags)
+                        metadata_dict[line_source_id]['New Tags'] = tags_arr
+            return metadata_csv, metadata_dict
+        else:
+            return metadata_csv, metadata_dict
```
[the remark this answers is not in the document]
kollivier: Also, the last line of the `if` and `else` blocks are the same, meaning you can move that line outside of the `if` block entirely and have it always run.

ricecooker/classes/nodes.py
```diff
@@ -212,6 +216,50 @@ def get_json_tree(self):
...
+        is_channel = isinstance(self, ChannelNode)
+ 
+        if not is_channel:
+            # Build out tag string
+            current_level = self.title
+            # prevent list from being passed by reference. Create new instance of list
+            structure_list = list(topic_structure)
+            # if len(structure_list) < 1:
+            #     structure_list = [current_level]
+            # else:
```
kollivier, line 232: It's a good idea to just delete code instead of commenting it out once you know you won't need it. Otherwise, people reading the code may wonder if you intended to add it back later.

ricecooker/classes/nodes.py
```diff
@@ -212,6 +216,50 @@ def get_json_tree(self):
 
         return tree
 
+
+    def save_channel_children_to_csv(self, metadata_csv, topic_structure = list()):
```
kollivier, line 220: Now that the reference error is fixed, you can just rename `topic_structure` to `structure_list`, and remove all references to `topic_structure`.

ricecooker/classes/nodes.py
kollivier: I'm still seeing the for loop inside both the if and else blocks. The reason we want to avoid having code that does the same thing in each block is that any time we change one we must remember to also change the other.

ricecooker/chefs.py
kollivier: Also, this function doesn't appear to use `metadata_csv` at all, so we can remove it.

ricecooker/classes/nodes.py
richard-dinh: Sorry about that. Changed to only one for loop

ricecooker/chefs.py
richard-dinh: Got it. Updated to use exist_ok = True

ricecooker/chefs.py
richard-dinh: You are right. Function no longer takes any arguments

ricecooker/chefs.py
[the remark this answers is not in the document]
richard-dinh: Removed else statement and moved outside if clause

ricecooker/classes/nodes.py
richard-dinh: Sorry about that. Comment has been removed

ricecooker/classes/nodes.py
```diff
@@ -226,14 +226,10 @@ def save_channel_children_to_csv(self, metadata_csv, topic_structure = list()):
             # Build out tag string
             current_level = self.title
             # prevent list from being passed by reference. Create new instance of list
-            structure_list = list(topic_structure)
-            # if len(structure_list) < 1:
-            #     structure_list = [current_level]
-            # else:
+            structure_list = list(structure_list)
```
kollivier, line 229: This line can completely be removed.

ricecooker/chefs.py
kollivier: The above issues all still appear in the latest version of the code.

ricecooker/chefs.py
```diff
@@ -281,10 +283,55 @@ def save_channel_tree_as_json(self, channel):
...
+                if line_new_title != '' or line_new_description != '' or line_new_tags != '':
+                    metadata_dict[line_source_id] = {}
+                    if line_new_title != '':
+                        metadata_dict[line_source_id]['New Title'] = line_new_title
+                    if line_new_description != '':
+                        metadata_dict[line_source_id]['New Description'] = line_new_description
+                    if line_new_tags != '':
+                        tags_arr = re.split(',| ,', line_new_tags)
+                        metadata_dict[line_source_id]['New Tags'] = tags_arr
+        return metadata_csv, metadata_dict
```
kollivier, line 316: We are still returning `metadata_csv` from this function and passing it into other functions, even though those functions do not access or use it.

ricecooker/classes/nodes.py
richard-dinh: I tried printing the structure list out with this line removed and it was printing out incorrectly. Seems that it was still passing in by reference

Capture

ricecooker/chefs.py
richard-dinh: I am a little confused here. the `save_modifications` function does need `metadata_csv`. If no csv is passed in, then does not run but if there is a csv with any updated fields, then it does run.

ricecooker/classes/nodes.py
kollivier: Ah, thanks, so it looks like what I read about `list()` as a function arg creating a new list each time was false. So let's make sure that any list or dict objects passed in default to `None` and that we set it when it is `None` the top of the function, like so:

```
if structure_list is None:
    structure_list = []
```

ricecooker/chefs.py
kollivier: Yes, but what does it need the `metadata_csv` object for? It checks that it exists, but does nothing with it. If the `metadata_csv` file is `None`, then `metadata_dict` will also be `{}`, therefore the two checks are redundant, and you only need the `metadata_dict` check before trying to apply the modifications to the nodes.

richard-dinh: Thank you for the explanation, csv has been removed

ricecooker/classes/nodes.py
richard-dinh: updated list to be named as `branch_structure_list `for now as it is keeping track of the structure of each branch. Can name it to something different in the future

ricecooker/chefs.py
richard-dinh: Sorry about that. Changed function name to apply_modifications

kollivier (approved):
Thanks for all your hard work on this! LGTM! :)

PR #299, opened 2020-12-10
Updated open channel url to match new Studio routes

jayoshih:


kollivier:
What will happen if we have URLs in the older format? Since we won't expect all chefs to be using the latest ricecooker, people will be passing around the old links for a while.

jayoshih:
It looks like it will auto redirect to the expected url

PR #301, opened 2021-01-19
Docs: Fixing syntax in PDF splitting example

bkarnow:
An additional pair of brackets is required around the `DocumentNode(...)`, otherwise `TypeError: 'DocumentFile' object is not iterable` will be thrown.


kollivier:
Sorry, seems I overlooked this? Fix looks right to me, any reason for closing it out?

bkarnow:
Oops, I think I accidentally killed the branch for this when I was trying to tidy my local git folder, I'll resubmit it soon.

PR #305, opened 2021-02-16
Adds webm formats and tests

sairina:
### Summary
Added webm as an allowed format for `VideoFile` class so that webm files do not get converted to mp4s, and tests for low- and high-resolution webm videos.

### DIFF
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -232,7 +232,7 @@ class Node(object):
                 new_tags = ','.join(new_tags)
 
             record = [
-                self.source_id, 
+                self.source_id,
                 structure_string,
                 self.title,
                 new_title,        # New Title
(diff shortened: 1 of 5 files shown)


ricecooker/classes/nodes.py
```diff
@@ -607,7 +607,8 @@ class VideoNode(ContentNode):
             files ([<File>]): list of file objects for node (optional)
     """
     kind = content_kinds.VIDEO
-    required_file_format = file_formats.MP4
```
rtibbles, line 610: Although it makes almost no difference because the validate function that used to use this has been commented out for some reason, I think it might be a neater convention to allow `required_file_format` to be either a string or a tuple, so I would changes this instead to:
```
required_file_format = (file_formats.MP4, file_formats.WEBM)
```

When we reinstate the validate function, we should then allow it to handle both strings and tuples as this property.

rtibbles (changes requested):
One thing I think might need to change - otherwise it looks good to go!

rtibbles (approved):

PR #307, opened 2021-02-23
sort_topic_nodes added to TreeNode

richard-dinh:
@kollivier  Creating PR due to request: https://github.com/learningequality/sushi-chef-chicken-and-chips/pull/1#discussion_r579429868

### DIFF
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -438,7 +438,7 @@ class TreeNode(Node):
             convert = lambda text: int(text) if text.isdigit() else text.lower() 
             key = lambda key: [ convert(re.sub(r'[^A-Za-z0-9]+', '', c.replace('&', 'and'))) for c in re.split('([0-9]+)', key.title) ]
         self.children = sorted(self.children, key = key, reverse = reverse)
-        return self
+        return self.children
 
     def to_dict(self):
         """ to_dict: puts Topic or Content node data into the format that Kolibri Studio expects
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -427,6 +427,26 @@ class TreeNode(Node):
 
         super(TreeNode, self).truncate_fields()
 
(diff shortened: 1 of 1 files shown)


ricecooker/classes/nodes.py
```diff
@@ -427,6 +427,26 @@ def truncate_fields(self):
 
         super(TreeNode, self).truncate_fields()
 
+    def sort_topic_nodes(self, channel, key = None, reverse = False):
```
kollivier, line 430: Please name this `sort_child_nodes` and remove the `channel` argument. We should sort `self.children`.

ricecooker/classes/nodes.py
```diff
@@ -427,6 +427,26 @@ def truncate_fields(self):
...
+        :param reverse: A Boolean. False will sort ascending, True will sort descending. False by default
+        :return: Sorted channel
+        """
+        # default natural sorting
+        if not key:
+            convert = lambda text: int(text) if text.isdigit() else text.lower() 
+            key = lambda key: [ convert(re.sub(r'[^A-Za-z0-9]+', '', c.replace('&', 'and'))) for c in re.split('([0-9]+)', key.title) ]
+        try:
+            channel.children = sorted(channel.children, key = key, reverse = reverse)
+        except Exception as e:
```
kollivier, line 444: Please remove the exception part. If the developer wants the default sorting, they can write fallback code themselves.

kollivier (changes requested):
Thanks! Needs a couple tweaks but otherwise looks good!

ricecooker/classes/nodes.py
kollivier: Also make sure to update the function doc strings to remove references to channel.

richard-dinh: changes have been made

ricecooker/classes/nodes.py
richard-dinh: removed exception

kollivier:
LGTM, thanks!

PR #308, opened 2021-02-25
Sqlite3 db export

kollivier:
Just some work I did as part of "passion project holidays." ;) Posting in case it's of use for any current projects or use cases.

Basically, it enables direct sqlite3 db exports of ricecooker channels. It does this by using a modified version of the Studio publishing code (found in `learningequality/kolibri-content-tools`). 

It also makes SushiChefs usable as library code, by removing the hard `sys.exit` calls on error and raising exceptions instead. This resolves some issues, like the need on many chefs to manually run them one language at a time because each time you create a SushiChef class it will exit at the end of its run.


rtibbles:
Closing for now, we can mine the code for useful updates in future!

PR #309, opened 2021-03-02
Remove mention of `--deploy` and `--publish` flags.

bkarnow:
Per discussion with Kevin, these are deprecated.


bkarnow:
Whoops, missed that this was merged into develop!

Issue #311, opened 2021-03-02
index_path - no such file or directory exists


richard-dinh:
* ricecooker version: most recent develop branch
* Python version: 3.7.9
* Operating System: Windows

### Description

Line with issue: [(https://github.com/learningequality/ricecooker/blob/develop/ricecooker/utils/downloader.py#L818)](https://github.com/learningequality/ricecooker/blob/develop/ricecooker/utils/downloader.py#L818)
When running the [Sciensation](https://github.com/learningequality/sushi-chef-sciensation) sushi chef script breaks down when it hits the above line and gives a `Errno 2`.

![error](https://user-images.githubusercontent.com/58611530/109726500-b3018000-7b67-11eb-96bc-e1f15e356e48.PNG)

This line was added in the following [commit](https://github.com/learningequality/ricecooker/commit/ad79ef3150e5beee6379795412abb5bf144de7eb)
### What I Did

```
Run Sciensation Chef
```

kollivier:
@richard-dinh Oops, sorry about that! Fixing it should just be a matter of calling `os.makedirs` on `os.path.dirname(os.path.join(temp_dir, index_path))` with `exist_ok=True` before trying to write the file.

richard-dinh:
@kollivier Thank you for the help! I added that in and noticed that issue, however there is no `index.html` at the root of zip folder. Was this intended? Was this [line](https://github.com/learningequality/ricecooker/commit/ad79ef3150e5beee6379795412abb5bf144de7eb#diff-58863fc4828722eb55772876b30ce481ba87e6c422b15aca36c365372b6c8df9L801) adding `index.html` to the root?

richard-dinh:
Is this [line](https://github.com/learningequality/ricecooker/commit/ad79ef3150e5beee6379795412abb5bf144de7eb#diff-58863fc4828722eb55772876b30ce481ba87e6c422b15aca36c365372b6c8df9R745) meant to be `if self.relative_links`? If so, would we still need to call `os.makedirs` since we do not want to make a directory at the root for `index.html`?

kollivier:
@richard-dinh No, that line should be `if not self.relative_links` as it is now.  When `self.relative_links` is True, we dynamically create a root `index.html` file that simply redirects to the page at `index_path`. That is the `entrypoint` that you noticed earlier.

richard-dinh:
[...]

PR #322, opened 2021-03-24
create directory for index_path

richard-dinh:
Fix for [issue](https://github.com/learningequality/ricecooker/issues/311)

### DIFF
--- a/ricecooker/utils/downloader.py
+++ b/ricecooker/utils/downloader.py
@@ -815,7 +815,7 @@ class ArchiveDownloader:
 
         index_path = self.get_relative_index_path(url)
 
-        os.makedirs(os.path.dirname(os.path.join(temp_dir, index_path)))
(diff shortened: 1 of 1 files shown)


ricecooker/utils/downloader.py
```diff
@@ -815,6 +815,7 @@ def create_zip_dir_for_page(self, url):
 
         index_path = self.get_relative_index_path(url)
 
+        os.makedirs(os.path.dirname(os.path.join(temp_dir, index_path)))
```
rtibbles, line 818: Should call `os.makedirs` with the `exist_ok=True` kwarg (https://docs.python.org/3/library/os.html#os.makedirs) so that it doesn't fail if the directory already exists.

rtibbles (changes requested):
One small tweak.

ricecooker/utils/downloader.py
richard-dinh: added!

rtibbles (approved):

PR #324, opened 2021-03-29
ricecooker.utils.pdf.PDFParser#get_toc reimplemented to use pdftk

### LINKED ISSUE #312: PDF Files created with wkhtmltopdf cannot have their outline parsed
### Description

Because of an extant bug in PyPDF2 https://github.com/mstamy2/PyPDF2/issues/193 trying to read the outline for a file generated in wkhtmltopdf results in an error. This means that large PDF files generated in this way cannot be split up into smaller PDFs.

### Potential resolutions

1. The issue contains a way to bypass this issue by patching the library - but it's not clear whether it actually fixes the problem, or just prevents a blocking error. I suspect that it just lets the PDF to continue to be parsed, but that we would not get the outline data - which we need in order to do the by chapter and subchapter splitting that we are using PyPDF2 for.

2. Doing a more systematic fix on this would potentially involve forking and diving deeper into PyPDF2, as there appears to be no current maintenance of the library.

3. There is another Python library https://github.com/pdfminer/pdfminer.six that is actively maintained - but it appears that it only extracts layout metadata. In order to use this, we would have to migrate the outline reading code to use pdfminer.six, but still continue to use PyPDF2 for the actual page splitting.

4. Another alternative is to use the pdftk binary https://www.pdflabs.com/tools/pdftk-server/ - while there is a Python wrapper library https://github.com/revolunet/pypdftk for this, it has a very limited API and does not expose the tools we need. However, pdftk is able to read the entire outline and dump all the relevant information to JSON, which we could then read back into Python. We could then also use pdftk to do the splitting as well.

In terms of ongoing maintainability, I think option 4 is probably the most promising bet, as we are leveraging a widely used open source binary to handle the complex business of PDF parsing. One possible upside is that whereas PyPDF2 failed to preserve accessibility focused aspects of PDFs during the splitting, pdftk might do a better job of this.

We would have to add good graceful failure and the ability to configure the path to the pdftk binary in order to leverage this.

To move forward with this strategy, we would have to remove PyPDF2 and replace operations with subprocess calls to operations in pdftk: https://www.pdflabs.com/docs/pdftk-man-page/

* The [get_toc](https://github.com/learningequality/ricecooker/blob/master/ricecooker/utils/pdf.py#L75) method would now leverage the pdftk `dump_data` command. Writing the JSON out to a temporary file in a subprocess, reading that JSON back into Python and then cleaning up the temporary file.
* The [split_chapters](https://github.com/learningequality/ricecooker/blob/master/ricecooker/utils/pdf.py#L155) and [split_subchapters](https://github.com/learningequality/ricecooker/blob/master/ricecooker/utils/pdf.py#L173) methods would now use the pdftk `cat` command and pass the page ranges parsed from `get_toc` to generate the new file.

Because of deferring to pdftk for the PDF processing, most of the other logic in the `PDFParser` class could be removed.

nucleogenesis:
Fixes #312 

WIP - Need to make a test that fails the old `get_toc` implementation (temporarily renamed to `get_old_toc`) but passes the new one to be sure that I'm addressing the issue noted about PDFs created with `wkhtmltopdf`.

Currently:

- Newly implemented `get_toc` passes existing test suites.
- `get_toc` uses `NamedTemporaryFile` and hopefully in a cross-OS compatible way (I only accessed it using the `.name` property once and close the file at the end anyway).
- I didn't see the need to update any other methods since the others mentioned by @rtibbles in #312 call `get_toc` for their data.

### DIFF
--- a/ricecooker/utils/pdf.py
+++ b/ricecooker/utils/pdf.py
@@ -1,36 +1,42 @@
 import os
 import subprocess
 
-from tempfile import NamedTemporaryFile
-from PyPDF2 import PdfFileWriter, PdfFileReader
-from PyPDF2.generic import Destination, NullObject
-from PyPDF2.utils import PdfReadError
+from tempfile import mkstemp
 from ricecooker.utils.downloader import read
 
 
-class CustomDestination(Destination):
-    def __init__(self, title, page, typ, *args):
-        try:
-            super(CustomDestination, self).__init__(title, page, typ, *args)
-        except PdfReadError:
-            pass
+# pdftk output keys for parsing control simplicity
+BOOKMARK_BEGIN = "BookmarkBegin"
+LAST_BOOKMARK_BEGIN = "PageMediaBegin"
+TITLE = "BookmarkTitle"
+LEVEL = "BookmarkLevel"
+PAGE_START = "BookmarkPageNumber"
+NUM_OF_PAGES = "NumberOfPages"
 
-class CustomPDFReader(PdfFileReader):
-    def _buildDestination(self, title, array):
-        page, typ = array[0:2]
-        array = array[2:]
-        return CustomDestination(title, page, typ, *array)
+
+def get_pdftk_line_value(line):
+    """
+    Returns everything right of the : in the given line. Useful for getting the value
+    for a key in pdftk dump_data* output
+    :param line:string
+    :return:string
+    """
+    # Get everything right of the : sans spaces and newlines and quotes
+    return line.split(":")[-1].strip(" \n")
 
 
 class PDFParser(object):
     """
     Helper class for extracting table of contents and splitting PDFs into chapters.
     """
-    path = None       # Local path to source PDF document that will be processed
+
+    path = None  # Local path to source PDF document that will be processed
 
     def __init__(self, source_path, directory="downloads"):
         self.directory = directory
         self.source_path = source_path
+        self.toc_subchapters = None
+        self.toc = None
 
     def __enter__(self):
         """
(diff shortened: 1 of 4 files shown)


ricecooker/utils/pdf.py
```diff
@@ -71,8 +73,100 @@ def check_path(self):
         if not self.path:
             raise ValueError("self.path not found; call `open` first")
 
-
     def get_toc(self, subchapters=False):
+        temp_file = NamedTemporaryFile()
```
rtibbles, line 77: Need to use `mkstemp` here - on windows trying to access a `NamedTemporaryFile` by referencing its name will give you a permissions error.

Also then need to explicitly close the file handle returned by mkstemp to ensure proper cleanup.

See here for an example: https://github.com/learningequality/ricecooker/blob/8cbdf5968457db8ea5d6d71f9d71f690bf8f0ad4/ricecooker/classes/files.py#L776

rtibbles (commented):
One update to start with - my suggestion in the issue was just to bite the bullet and use pdftk for all of our PDF operations, so we could drop reliance on the unmaintained Python library.

ricecooker/utils/pdf.py
nucleogenesis: > One update to start with - my suggestion in the issue was just to bite the bullet and use pdftk for all of our PDF operations, so we could drop reliance on the unmaintained Python library.

There is one issue here where we'll break the PDFParser API which I mentioned in my comment on the linked issue: https://github.com/learningequality/ricecooker/issues/312 - the `PDFParser.pdf` field is (in code) a public property. [It isn't mentioned in the docs though](https://ricecooker.readthedocs.io/en/latest/pdfutils.html). I'm okay with moving forward this way but feel it just needs to be noted that some peoples' chefs might break if they've managed to rely on that property at all - which they might because it is a PyPDF2 instance.

Will continue on with fully wrapping pdftk - but as we are effectively not doing anything to improve the software, I also suggested we consider just catching the error and informing the user of the limitations of PDFParser - which is that we cannot help them if their PDF doesn't have the proper metadata. This would be equally robust a solution if it can be worked out and ostensibly simpler. 

We have to do something like this anyway if we use pdftk.

Happy to go either way just wanted to hear your thoughts on the above.

nucleogenesis: Also - its worth noting that the change I made is basically removing the meat of PyPDF2.

What remains is some clean-up and documentation updates to finalize the removal of PyPDF2. Also - I need to catch and raise a meaningful error message when we get a PDF without an outline.

rtibbles: Yep - I think catching errors and giving meaningful error messages is an improvement :)

rtibbles:
I also recently came across this: https://github.com/jorisschellekens/borb - not sure if it would help or not.

nucleogenesis:
That will probably be necessary. I think I know what the issue is - the latest Mac OS (10.15ish) doesn't run 32-bit applications and pdftk only targets back to 10.9.

I can give borb a look as an alternative for sure

nucleogenesis:
re: `borb` - I played with it a bit this AM and here are some notes for when returning here.

load a file like:

```python
with open('path/to/file', 'rb') as f:
    doc = borb.pdf.pdf.PDF.loads(f)
    outlines = doc["XRef"]["Trailer"]["Root"]["Outlines"] # is where the outlines live
    
    from borb.io.read.types import Name, String, Dictionary
    
    # when you look into outlines, you'll see a dictionary of objects of thoes Name, String, Dictionary types
    # which means you have to look at the source to know how to get the values from them like String you call
    # __str__() on it... it's a mess. This should be enough of a head start upon returning.
```

nucleogenesis:
I played a bit more with `borb` and it turns out that the metadata it does pull from a PDF is currently incomplete - only extracting a subset of the outlines metadata that both PyPDF2 and pdftk extracted successfully.

nucleogenesis:
[cpdf](https://github.com/coherentgraphics/cpdf-binaries) is free for non-commercial use and does everything we want.

It was a pretty quick shift - but now tox is mad about subprocesses. I wanted to capture stderr during dev so I used subprocess.run() and didn't write to a tmp file like before. Not sure if this is the cause, but the error is coming up in subprocess.Popen - so despite finding a tool that works for our needs very well, back at square one with failing tests on the binaries.

Will come back to this again at some point.

nucleogenesis:
Stale - closing for now

PR #329, opened 2021-04-19
Ensures that custom navigation is properly set as 'options' on extra_fields

### LINKED ISSUE #328: Ensure topic/channel extra fields are added as 'options' in their extra_fields property to make sure options for topics are set during publishing
It appears that currently extra fields on topics are not annotated onto the exported topics during publishing.

This is required in order to properly export the modality extra field and expose `CUSTOM_NAVIGATION` modality to Kolibri for determining if a topic has this modality.

rtibbles:
Fixes #328

Ensures that the custom navigation modality is set in the options of extra fields, so it gets directly exported in publishing.

This ensures it gets set in `options` on the kolibri exported contentnode.

Screenshot of published channel:

Screenshot from 2021-04-19 11-49-01

### DIFF
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -1069,7 +1069,9 @@ class CustomNavigationNode(ContentNode):
     required_file_format = file_formats.HTML5
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


marcellamaki (approved):
This looks good to me

rtibbles:
Looks like I broke my own tests, so I need to fix these!

PR #339, opened 2021-07-13
Direct GCS Upload

rtibbles:
Implements direct upload to Google Cloud Storage

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -381,6 +381,18 @@ class File(object):
 
     def get_filename(self):
         return self.filename or self.process_file()
+    
+    @property
+    def checksum(self):
+        return self.get_filename().split(".")[0]
+    
+    @property
+    def extension(self):
+        return self.get_filename().split(".")[1]
+    
+    @property
+    def size(self):
+        return os.path.getsize(config.get_storage_path(self.get_filename()))
 
     def truncate_fields(self):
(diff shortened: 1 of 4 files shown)


dylanmccall:
This worked well from my own testing, with the obvious limitation that it turns out 2 GB files aren't actually supported at the moment :)

@rtibbles I made another branch over here which is this branch + a change that better handles multipart file uploads: https://github.com/learningequality/ricecooker/compare/develop...dylanmccall:add-multipart-encoder--gcs_upload. The `requests` library on its own sticks everything in memory, which breaks once we add SSL. (I guess the C library doesn't like 2 GB strings). Using `MultipartEncoder` from `requests_toolbelt`, we can upload huge files no problem.

rtibbles:
@dylanmccall I merged your PR to mine, but this has broken uploading now. I don't think the MultiPart upload is playing well with GCS, as with the current URL it fails MD5 checks.

rtibbles:
Going to remove the multipart encoder piece for now. We will need to update the file URL generation for resumable uploads.

rtibbles:
@dylanmccall I think the SSL interaction must be doing something strange here, as the recommendation from requests is to pass the file object as data in order to do a streamed upload, precisely as we are doing here: https://docs.python-requests.org/en/master/user/advanced/#streaming-uploads

PR #344, opened 2021-09-01
Chefs allow presets for files

AtKristijan:


tests/test_tree.py
```diff
@@ -159,12 +196,75 @@ def test_validate_tree(tree, invalid_tree, invalid_tree_2):
...
+    topic_node["children"].append(audio_node)
+    parent_node = build_tree_from_json(channel, [topic_node])
+    topic_node = parent_node.children[0]
+    html5_node = topic_node.children[0]
+    assert parent_node.validate_tree()
+    assert parent_node
+    assert parent_node.children[0]
+    assert topic_node.kind == 'topic'
+    assert len(html5_node.files) == 3
+    assert html5_node.files[2].get_preset() == format_presets.AUDIO_DEPENDENCY
```
rtibbles, line 267: This is the right test, so you have definitely fixed the issue!

rtibbles (changes requested):
The issue seems definitely to be fixed here - just need to remove the changes adding presets to content nodes which we don't want.

rtibbles:
The test failure seems to be due to some flakiness of some tests you didn't touch - if it persists, we can file an issue and merge anyway.

rtibbles (approved):
Great cleanup, thanks!

PR #347, opened 2021-12-06
Add modality to ExerciseNode and pre-commit linting

AtKristijan:
Adding modality to the node and set to be QUIZ


ricecooker/classes/nodes.py
```diff
@@ -887,7 +892,7 @@ def __init__(self, source_id, title, license, questions=None, exercise_data=None
             'mastery_model': exercise_data.get('mastery_model', exercises.M_OF_N),
             'randomize': exercise_data.get('randomize', True),
         })
-
+        kwargs["extra_fields"]["options"].update({'modality': "QUIZ"})
```
rtibbles, line 895: This does not meet the spec as set out in the issue:
> This should be a descendant of the ExerciseNode class, but with the addition that the modality in the options key of extra_fields should be set to QUIZ.

You have just added this to the ExerciseNode here, meaning that all ExerciseNodes will now have this set.

The requirement is to create a new QuizNode that is a descendant of ExerciseNode that sets this.

rtibbles (changes requested):

rtibbles:
Closing in favour of #349

PR #348, opened 2021-12-06
Fix pytest windows

AtKristijan:
Change poppler link for windows zip.


rtibbles (approved):
Tests are now running! We need to do some follow up to fix the remaining tests, but this PR does its job.

PR #351, opened 2021-12-21
Pre commit for ricecooker

AtKristijan:
Precommit for ricecooker


.pre-commit-config.yaml
```diff
@@ -1,19 +1,31 @@
...
+        hooks:
+            -   id: black
+                types_or: [ python, pyi ]
+                exclude: '^.+?\.template$'
+    -   repo: https://github.com/pycqa/flake8
+        rev: 4.0.1
+        hooks:
+            -   id: flake8
+                exclude: '^.+?\.template$'
+    -   repo: git://github.com/pre-commit/pre-commit-hooks
```
rtibbles, line 13:
```suggestion
    -   repo: https://github.com/pre-commit/pre-commit-hooks
```

A recent change in github security means that the unauthenticated `git://` protocol URLs will no longer work. I am making similar changes in our other repos: https://github.com/learningequality/kolibri/pull/8987

.pre-commit-config.yaml
```diff
@@ -1,19 +1,31 @@
...
+            -   id: check-yaml
+                exclude: '^.+?\.template$'
+            -   id: check-added-large-files
+                exclude: '^.+?\.template$'
+            -   id: debug-statements
+                exclude: '^.+?\.template$'
+            -   id: end-of-file-fixer
+                exclude: '^.+?(\.json|\.template)$'
+
+    -   repo: git://github.com/asottile/reorder_python_imports
```
rtibbles, line 27:
```suggestion
    -   repo: https://github.com/asottile/reorder_python_imports
```

rtibbles (commented):
No concerns here - there's a bit more cleanup to do with the linting from the flake-8 checks though, they are somewhat extensive, so we could break them up into follow up issues.

.pre-commit-config.yaml
AtKristijan: changed

.pre-commit-config.yaml
AtKristijan: changed

rtibbles:
Going to merge this so we can iterate on getting to passing tests again.

PR #352, opened 2022-01-13
extract video duration.

AtKristijan:
Work in progress.

Add function for extract duration of video.


ricecooker/utils/videos.py
```diff
@@ -48,13 +49,24 @@ def extract_thumbnail_from_video(fpath_in, fpath_out, overwrite=False):
...
+                   "-nostats",
+                   '-ss', str(midpoint), '-vframes', '1', '-q:v', '2', "-loglevel", "panic", str(fpath_out)]
         subprocess.check_output(command, stderr=subprocess.STDOUT)
     except subprocess.CalledProcessError as e:
         raise ThumbnailGenerationError("{}: {}".format(e, e.output))
 
 
+def extract_duration_of_video(fpath_in):
+    try:
+        result = subprocess.check_output(['ffprobe', '-v', 'error', '-show_entries', 'format=duration', '-of',
```
rtibbles, line 62: Would this same command work for an audio file too? In which case, we can make the function name more generic like `extract_duration_of_media`.

AtKristijan: Yes it can be change.  This work for both audio and video. I tested with both.

rtibbles: That would be good - can also add the duration pieces you've added to video file classes to audio files too.

rtibbles:
I just merged your linting PR, looks like there are now merge conflicts with develop - so might want to rebase here.

ricecooker/classes/files.py
```diff
@@ -551,10 +552,14 @@ class AudioFile(DownloadFile):
     default_ext = file_formats.MP3
     allowed_formats = [file_formats.MP3]
     is_primary = True
+    duration = None
 
     def get_preset(self):
         return self.preset or format_presets.AUDIO
 
+    def get_duration(self):
```
rtibbles, line 560: I think it might be simpler to not add the `get_duration` methods, and just add `process_file` to the `AudioFile` class. As it derives from `DownloadFile` can call super on the `process_file` method, then set the duration on the returned filename.

ricecooker/utils/videos.py
rtibbles: Still need to update the method name here.

ricecooker/classes/nodes.py
```diff
@@ -863,6 +867,7 @@ def validate(self):
         from .files import AudioFile
 
         try:
+            self.duration = audio_files[0].get_duration()
```
rtibbles, line 870: I don't see where `audio_files` is defined here? Did you forget to add the list comprehension above?

rtibbles (changes requested):
One thing that looks broken, another thing that I think could be improved.

ricecooker/classes/nodes.py
AtKristijan: When i merge the code it was delete by mistake.

ricecooker/utils/videos.py
AtKristijan: Name changed

ricecooker/classes/files.py
AtKristijan: Changed this too

ricecooker/classes/files.py
```diff
@@ -735,6 +751,8 @@ def process_file(self):
 
         return self.filename
 
+    def get_duration(self):
```
rtibbles, line 754: I think to simplify things, we can just remove these `get_duration` methods from the file classes - they are never referred to anywhere, and just setting the `duration` property should be sufficient.

rtibbles (approved):

PR #355, opened 2022-01-31
fix: tag length validation

vkWeb:
This PR is in reference to https://github.com/learningequality/studio/issues/3296. The tag character length check was expecting them to be less than or equal to 50 chars but in Kolibri we enforce a 30 char limit. This PR fixes that validation.


bjester (approved):

bjester:
Merging as there are existing issues with linting and python tests that are unrelated to this change

PR #362, opened 2022-02-21
Fix pytest and tox

AtKristijan:
Work in progress

Fixing unit test for tox to work properly.


ricecooker/classes/files.py
```diff
@@ -559,8 +559,8 @@ def get_preset(self):
 
     def process_file(self):
         self.filename = super(AudioFile, self).process_file()
-        self.duration = extract_duration_of_media(self.filename)
-
+        self.duration = extract_duration_of_media(self.path)
```
rtibbles, line 562: I think we actually want to use: `config.get_storage_path(self.filename)` here, instead of `self.path` - for a file downloaded from the Internet, `self.path` will be a URL.

See here for example: https://github.com/learningequality/ricecooker/blob/33ca5c2cfdbae1057e2880f17790e55da3f79048/ricecooker/classes/files.py#L1193

ricecooker/classes/files.py
```diff
@@ -679,7 +679,7 @@ def process_file(self):
                         self.filename, self.ffmpeg_settings
                     )
                     config.LOGGER.info("\t--- Compressed {}".format(self.filename))
-            self.duration = extract_duration_of_media(self.filename)
+            self.duration = extract_duration_of_media(self.path)
```
rtibbles, line 682: Likewise here.

ricecooker/classes/files.py
```diff
@@ -736,7 +736,7 @@ def process_file(self):
             if self.filename and config.COMPRESS:
                 self.filename = compress_video_file(self.filename, {})
                 config.LOGGER.info("\t--- Compressed {}".format(self.filename))
-            self.duration = extract_duration_of_media(self.filename)
+            self.duration = extract_duration_of_media(self.path)
```
rtibbles, line 739: And here.

ricecooker/classes/files.py
```diff
@@ -1030,15 +1030,15 @@ class _ExerciseGraphieFile(DownloadFile):
...
+            path.split(os.path.sep)[-1].split(".")[0]
         )
         super(_ExerciseGraphieFile, self).__init__(path, **kwargs)
 
     def get_preset(self):
         return self.preset or format_presets.EXERCISE_GRAPHIE
 
     def get_replacement_str(self):
-        return self.path.split("/")[-1].split(".")[0] or self.path
+        return self.path.split('/')[-1].split(".")[0] or self.path
```
rtibbles, line 1041: This doesn't seem to have actually changed? I think maybe we could just return `self.original_filename` here, to avoid reduplicating the logic?

ricecooker/config.py
```diff
@@ -340,7 +340,8 @@ def get_storage_path(filename):
     Args: filename (str): Name of file to store
     Returns: string path to file
     """
-    directory = os.path.join(STORAGE_DIRECTORY, filename[0], filename[1])
+    # directory = os.path.join(STORAGE_DIRECTORY, filename[0], filename[1])
```
rtibbles, line 343: Should delete any commented out code.

rtibbles (commented):
Not seeing why that test is still failing in that way on Windows - but I would guess that it is pointing to the wrong path somehow, and that's what's giving the different checksum.

ricecooker/classes/files.py
AtKristijan: self.path is working fine i do not think that we should change this.

ricecooker/classes/files.py
AtKristijan: Nothing is changed here

ricecooker/classes/files.py
AtKristijan: If we are using this with get_storage_path(filename ) then we need extra check in the function because no all test got filename, some test where file does not exists will fail.

rtibbles:
Squashing and merging this to get this in so we can iterate further.

Issue #363, opened 2022-03-14
Fix Youtube_DL bug with download


AtKristijan:
* ricecooker version:
* Python version:
* Operating System:

### Description

When I tried to create new channel with chef script i need it to use ricecooker youtube_dl or WebVideoFile. The problem here is that the youtube_dl is crashing, it was old version and not updated. 
The other bug is with self.path because webVideoFile do not have that.

### What I Did

The solution that is possible is to replace youtube_dl with another fork of this library that got better maintenance and its yt_dlp with this its working but its slow. 

The second solution that need to be implemented it is to use pytube instead of youtube_dl. 
Tested all 3 libraries. The best results are get with pytube. Currently in my chef script using this pytube to download and to make the channel. Not using ricecooker for downloading because its slow.

PR #364, opened 2022-03-14
Issue 363 library problem

AtKristijan:
https://github.com/learningequality/ricecooker/issues/363

I change the library and change the bug with duration function.

### DIFF
--- a/ricecooker/utils/youtube.py
+++ b/ricecooker/utils/youtube.py
@@ -7,7 +7,9 @@ import time
 from datetime import datetime
 from enum import Enum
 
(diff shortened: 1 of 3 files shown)


ricecooker/utils/youtube.py
```diff
@@ -7,7 +7,9 @@
 from datetime import datetime
 from enum import Enum
 
-import youtube_dl
+# import yt_dlp
```
rtibbles, line 10:
```suggestion
```

rtibbles (approved):

PR #370, opened 2022-04-12
Update metadata labels.

### LINKED ISSUE #367: Add support for metadata labels to ricecooker
Ricecooker needs to support adding the new metadata labels to ContentNodes.

The following fields should be supported:
```
    grade_levels
    resource_types
    learning_activities
    accessibility_labels
    categories
    learner_needs
```

In addition, any data should be validated before being sent to Studio, ensuring that only valid labels in `le_utils` are sent.

They are all imported from le_utils.constants.labels with the following field to module mapping:
```
    grade_levels : levels
    resource_types : resource_type
    learning_activities : learning_activities
    accessibility_labels : accessibility_categories
    categories : subjects
    learner_needs : needs
```

For more information see https://github.com/learningequality/studio/issues/3356

AtKristijan:
Add new labels for metadata in ricecooker and validate them with le_utils

[fixes #367](https://github.com/learningequality/ricecooker/issues/367)

### DIFF
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -561,6 +561,13 @@ class TreeNode(Node):
             "copyright_holder": "",
             "questions": [],
             "extra_fields": json.dumps(self.extra_fields),
+            "grade_levels": None,
+            "resource_types": None,
+            "learning_activities": None,
+            "accessibility_categories": None,
+            "subjects": None,
+            "needs": None,
+
         }
 
     def validate(self):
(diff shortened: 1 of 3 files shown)


ricecooker/classes/nodes.py
```diff
@@ -9,14 +9,21 @@
...
 from .. import __version__
 from .. import config
 from ..exceptions import InvalidNodeException
 from .licenses import License
 
 MASTERY_MODELS = [id for id, name in exercises.MASTERY_MODELS]
 ROLES = [id for id, name in roles.choices]
+GRADE_LEVELS = [id for id, name in levels.choices]
+RESOURCE_TYPES = [id for id, name in resource_type.choices]
+ACCESSIBILITY_CATEGORIES = [id for id, name in accessibility_categories.choices]
```
rtibbles, line 23: Each of these label types has an explicit list of the available labels predefined in the le-utils module: https://github.com/learningequality/le-utils/blob/main/le_utils/constants/labels/accessibility_categories.py#L23 so can avoid doing these list comprehensions for these.

ricecooker/classes/nodes.py
```diff
@@ -691,6 +711,31 @@ def validate(self):
...
+            self.accessibility_labels in ACCESSIBILITY_CATEGORIES
+        ), "Assumption Failed: Accessibility label must be one of the following {}".format(ACCESSIBILITY_CATEGORIES)
+
+        assert (
+            self.categories in SUBJECTS
+        ), "Assumption Failed: Categories must be one of the following {}".format(SUBJECTS)
+
+        assert (
+            self.learner_needs in LEARNER_NEEDS
+        ), "Assumption Failed: Learner must be one of the following {}".format(LEARNER_NEEDS)
```
rtibbles, line 737: Should be `Learner Needs`

rtibbles (changes requested):
A good start - we also need to include these labels in the dict that is created to be sent to Studio.

rtibbles:
Can you ensure that this PR is rebased to latest develop? The linting and tests should all be passing.

rtibbles:
This PR is aimed at fixing [#367](https://github.com/learningequality/ricecooker/issues/367) - it does not fix [#3356](https://github.com/learningequality/studio/issues/3356), as that needs to be implemented on Studio to allow the sending of these new labels via the ricecooker specific endpoints.

ricecooker/classes/nodes.py
AtKristijan: Changed.

ricecooker/classes/nodes.py
AtKristijan: This is changed too

AtKristijan:
I checked and tried to rebase but it was last one. I also did pre-commit to check if everything is okay with linting

AtKristijan:
My local machine and github is giving different results for linting. Reorder is the problem.

ricecooker/classes/nodes.py
```diff
@@ -691,6 +710,44 @@ def validate(self):
         assert (
             self.role in ROLES
         ), "Assumption Failed: Role must be one of the following {}".format(ROLES)
+
+        assert (
+            self.grade_levels in levels.LEVELSLIST
```
rtibbles, line 715: Need to conditionalize these assertions:

First check `if self.grade_levels is not None:`

Then you need to check each of the values if it is a list, and see if they are in the `levels.LEVELSLIST`.

ricecooker/classes/nodes.py
```diff
@@ -691,6 +710,44 @@ def validate(self):
...
+        ), "Assumption Failed: Grade levels must be one of the following {}".format(
+            levels.LEVELSLIST
+        )
+
+        assert (
```
rtibbles, line 720: Same goes here and for all the other items.

ricecooker/classes/nodes.py
```diff
@@ -691,6 +710,44 @@ def validate(self):
...
+        ), "Assumption Failed: Categories must be one of the following {}".format(
+            subjects.SUBJECTSLIST
+        )
+
+        assert (
+            self.learner_needs in needs.NEEDSLIST
+        ), "Assumption Failed: Learner needs must be one of the following {}".format(
+            needs.NEEDSLIST
+        )
+
```
rtibbles, line 750: You also need to include all of this metadata in the data that we send to Studio - in the `to_dict` method of the TreeNode: https://github.com/learningequality/ricecooker/blob/develop/ricecooker/classes/nodes.py#L531 and also the `to_dict` method of the ContentNode: https://github.com/learningequality/ricecooker/blob/develop/ricecooker/classes/nodes.py#L706

rtibbles (changes requested):
Still some updates to do here.

I have had similar issues with the import sorting - it seems to be something to do with whether modules are available locally or not when the import sorting happens, but I've not found a good fix for it. Would be good to work out if you have any ideas.

rtibbles (approved):

rtibbles:
Hrm, not sure why but locally my pre-commit is now agreeing with the github action.

Issue #371, opened 2022-04-18
Content-disposition header is not read to determine filenames


rtibbles:
* ricecooker version: all

### Description

ricecooker currently only uses the file path (URL or local path) to determine what the extension of a file should be (unless coerced with a default file extension). However, many websites use a content disposition header to suggest an appropriate filename that should be used when downloading. Not appropriately reading these headers can result in misnamed files.

Something like this:
https://aviaryan.com/blog/gsoc/downloading-files-from-urls

should be implemented in ricecooker in order to facilitate better downloading.

rtibbles:
Fixed in #372

PR #372, opened 2022-04-29
Get ext content disposition

AtKristijan:
Finding extension from header

https://github.com/learningequality/ricecooker/issues/371


rtibbles (changes requested):
Let's do this without doubling the number of file downloads we do.

rtibbles (approved):

PR #380, opened 2022-06-28
Add a type of DownloadFile that can be split

dylanmccall:
This allows Kolibri to create content nodes from arbitrarily large source files for certain types of content.


rtibbles:
Closing in favour of #678

PR #393, opened 2022-09-16
Add workaround for bug in file upload endpoint.

rtibbles:


rtibbles:
Merging for immediate use.

PR #394, opened 2022-09-22
Default values for some metadata on ContentNode set to list()

### LINKED ISSUE #385: TreeNode to_dict is incompatible with Studio unstable branch
* ricecooker version: v0.7.0-beta6
* Python version: 3.7
* Operating System: Fedora Silverblue 36

### Description

I am creating a ricecooker script which makes use of the new structured metadata fields in the `develop` branch. This script creates a tree of TopicNode and VideoNode objects.

This is all working almost correctly.

The video nodes end up with attributes like `node.grade_levels = [levels.LOWER_PRIMARY]`, and the [`ContentNode.to_dict()`](https://github.com/learningequality/ricecooker/blob/aa857d102fc19a066931a8fe97e4d869e6d0d606/ricecooker/classes/nodes.py#L797-L832) function for each video node produces the expected schema, with something like `{…, "grade_levels": "wnarlxKo", …}`.

The topic nodes do not have the structured metadata attributes. However, the [`TreeNode.to_dict()`](https://github.com/learningequality/ricecooker/blob/aa857d102fc19a066931a8fe97e4d869e6d0d606/ricecooker/classes/nodes.py#L545-L578) function produces output like `{…, "grade_levels": null, …}`.

This is incompatible with the code in Studio which validates attributes for nodes: https://github.com/learningequality/studio/blob/269e7e0b677c569c3a68c0a30a0d0fa342f190c0/contentcuration/contentcuration/views/internal.py#L667-L675. If a structured metadata attribute is included in the body, it _must_ be a list.

The result is a failed upload, with an error like the following:

```
(319)add_nodes()
    318             ]
--> 319             for chunk in chunks:
    320                 payload_children = []

ipdb> response.content
b"['grade_levels must pass a list of values']"
ipdb> response.request.body
'{"root_id": "db6de12407f24f1c9e36752abd49ef2f", "content_data": [{"title": "Basketball", "language": "en", "description": "", "node_id": "8ed9e0083dc45373a385d2ab2d001132", "content_id": "ce34868e07ec5404a35de380ccd502dd", "source_domain": "379151f326fd5a0080473baf3c4c396e", "source_id": "basketball", "author": "", "aggregator": "", "provider": "", "files": [], "tags": [], "kind": "topic", "license": null, "license_description": null, "copyright_holder": "", "questions": [], "extra_fields": "{}", "grade_levels": null, "resource_types": null, "learning_activities": null, "accessibility_categories": null, "subjects": null, "needs": null}, {"title": "Soccer", "language": "en", "description": "", "node_id": "357293c10e8b59e5b383da07d62c6c90", "content_id": "030cbf9bbc8353bbbbb91d641b6318e9", "source_domain": "379151f326fd5a0080473baf3c4c396e", "source_id": "soccer", "author": "", "aggregator": "", "provider": "", "files": [], "tags": [], "kind": "topic", "license": null, "license_description": null, "copyright_holder": "", "questions": [], "extra_fields": "{}", "grade_levels": null, "resource_types": null, "learning_activities": null, "accessibility_categories": null, "subjects": null, "needs": null}, {"title": "Juggling", "language": "en", "description": "", "node_id": "a5f005393b7b5841a6b067a99779d7b5", "content_id": "ef04f09ef399570aba145495ddcce80a", "source_domain": "379151f326fd5a0080473baf3c4c396e", "source_id": "juggling", "author": "", "aggregator": "", "provider": "", "files": [], "tags": [], "kind": "topic", "license": null, "license_description": null, "copyright_holder": "", "questions": [], "extra_fields": "{}", "grade_levels": null, "resource_types": null, "learning_activities": null, "accessibility_categories": null, "subjects": null, "needs": null}]}'
```

nucleogenesis:
Fixes #385 

@AllanOXDi and I worked through this together.

We updated the default values for the metadata values in ContentNode to an empty list rather than None. This should resolve the reported issue.

To test this, we enumerated the keys that must default to a list and ensure a default ContentNode descendant is initialized with an empty list. This test logic is only added to the test function that tests Video content -- as all content types descend from ContentNode, this test should be sufficient coverage without duplicating the code across each of the functions.

The existing tests failed because of the improperly matched labels -- we changed the names in `nodes.py`. The tests check that a defined fixture is correct against a created instance -- so please double check this for us in case we misunderstood as the change may have needed to occur at the `conftest.py` definition of the fixtures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ test-all: clean-test ## run tests on every Python version with tox
 	tox
 
 integration-test:
-	python tests/test_chef_integration.py
+	CONTENTWORKSHOP_URL=https://hotfixes.studio.learningequality.org python tests/test_chef_integration.py
 
 coverage: ## check code coverage quickly with the default Python
 	pip install coverage pytest
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,9 @@ test: clean-test ## run tests quickly with the default Python
 test-all: clean-test ## run tests on every Python version with tox
 	tox
 
+integration-test:
+	python tests/test_chef_integration.py
+
 coverage: ## check code coverage quickly with the default Python
 	pip install coverage pytest
 	coverage run --source ricecooker -m pytest
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,12 @@ test-all: clean-test ## run tests on every Python version with tox
 	tox
 
 integration-test:
+	echo "Testing against hotfixes"
 	CONTENTWORKSHOP_URL=https://hotfixes.studio.learningequality.org python tests/test_chef_integration.py
+	echo "Testing against unstable"
+	CONTENTWORKSHOP_URL=https://unstable.studio.learningequality.org python tests/test_chef_integration.py
+	echo "Testing against production"
+	CONTENTWORKSHOP_URL=https://studio.learningequality.org python tests/test_chef_integration.py
 
 coverage: ## check code coverage quickly with the default Python
 	pip install coverage pytest
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -463,6 +463,12 @@ class TreeNode(Node):
         provider="",
         tags=None,
         domain_ns=None,
+        grade_levels=None,
+        resource_types=None,
+        learning_activities=None,
+        accessibility_labels=None,
+        categories=None,
+        learner_needs=None,
         **kwargs
     ):
         # Map parameters to model variables
@@ -477,6 +483,13 @@ class TreeNode(Node):
         self.questions = (
             self.questions if hasattr(self, "questions") else []
         )  # Needed for to_dict method
+
+        self.grade_levels = grade_levels or []
+        self.resource_types = resource_types or []
+        self.learning_activities = learning_activities or []
+        self.accessibility_labels = accessibility_labels or []
+        self.categories = categories or []
+        self.learner_needs = learner_needs or []
 
         super(TreeNode, self).__init__(title, **kwargs)
 
@@ -569,12 +582,12 @@ class TreeNode(Node):
             "copyright_holder": "",
             "questions": [],
             "extra_fields": json.dumps(self.extra_fields),
-            "grade_levels": None,
-            "resource_types": None,
-            "learning_activities": None,
-            "accessibility_categories": None,
-            "subjects": None,
-            "needs": None,
+            "grade_levels": self.grade_levels,
+            "resource_types": self.resource_types,
+            "learning_activities": self.learning_activities,
+            "accessibility_labels": self.accessibility_labels,
+            "categories": self.categories,
+            "learner_needs": self.learner_needs,
         }
 
     def validate(self):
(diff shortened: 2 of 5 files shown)


rtibbles (commented):
Code changes make sense to me. In conversation we came up with the idea of adding to the tutorial sushichef script to act as both documentation of the new functionality, and to act as an integration test.

nucleogenesis:
@rtibbles ready for re-review if you want to try to run the new test command

tests/test_chef_integration.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python
+from le_utils.constants import licenses
```
rtibbles, line 2: I was rather expecting us to just edit the existing tutorial chef: https://github.com/learningequality/ricecooker/blob/develop/examples/tutorial/sushichef.py and run that as an integration test, to avoid duplicaton.

nucleogenesis: The actual sushichef file has the template strings and if you run a chef with the template strings in the config it won't run.

Alternatively, we could remove that check on running the chef. 

In any case, I figured that something that is a test is liable to be extended with more robust testing scenarios that wouldn't be relevant to the tutorial chef.

ricecooker/classes/nodes.py
```diff
@@ -677,12 +677,12 @@ def __init__(
         role=roles.LEARNER,
         license_description=None,
         copyright_holder=None,
-        grade_levels=None,
-        resource_types=None,
-        learning_activities=None,
-        accessibility_labels=None,
-        categories=None,
-        learner_needs=None,
+        grade_levels=[],
```
rtibbles, line 680: I think I see the problem - which was with the original implementation, not the updates you've done here.

These kwargs need to be moved to the `TreeNode` implementation. Then, the updates you've made to the `to_dict` method, also need to be added there, as the moment it is always returning `None` for each of these, which means it is still breaking, and also that we can't set any of this metadata for topics either.

ricecooker/classes/nodes.py
```diff
@@ -677,12 +677,12 @@ def __init__(
...
+        resource_types=[],
+        learning_activities=[],
```
rtibbles, line 682: There's also another gotcha here - you shouldn't use mutable types as default arguments, because they get shared across invocations of the method.

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

So, these defaults should still be `None` and when you come to assign them you would do:

```
        self.grade_levels = grade_levels or []
        self.resource_types = resource_types or []
        self.learning_activities = learning_activities or []
        self.accessibility_labels = accessibility_labels or []
        self.categories = categories or []
        self.learner_needs = learner_needs or []
```

rtibbles (changes requested):
Things to update:

1. Move the kwargs and kwargs setting on the node from `ContentNode` `__init__` to the TreeNode `__init__`.
2. Don't use mutable types as a default for a kwarg.
3. Copy the implementation for grade_levels etc. in the `to_dict` method of `ContentNode` to the `TreeNode` implementation (it needs to be in both places until we make this better).

ricecooker/classes/nodes.py
nucleogenesis: > Python’s default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, ~Ruby~ _every reasonable programming language_)

:face_with_spiral_eyes:

Makefile
```diff
@@ -59,6 +59,9 @@ test: clean-test ## run tests quickly with the default Python
 test-all: clean-test ## run tests on every Python version with tox
 	tox
 
+integration-test:
+	CONTENTWORKSHOP_URL=https://hotfixes.studio.learningequality.org python tests/test_chef_integration.py
```
rtibbles, line 63: Wondering if we want to test this against all three branches, but not a blocker.

ricecooker/classes/nodes.py
```diff
@@ -823,12 +830,6 @@ def to_dict(self):
             "extra_fields": json.dumps(self.extra_fields),
             "role": self.role,
             "suggested_duration": self.suggested_duration,
-            "grade_levels": self.grade_levels,
-            "resource_types": self.resource_types,
-            "learning_activities": self.learning_activities,
-            "accessibility_categories": self.accessibility_labels,
-            "subjects": self.categories,
-            "needs": self.learner_needs,
```
rtibbles, line 831: These still need to be here - the `to_dict` method doesn't inherit currently, so it needs to repeat what `TreeNode` does.

rtibbles (changes requested):
Just one more thing...

PR #409, opened 2023-01-26
Compression updates

rtibbles:
* Adds compression for audio files
* Adds compression for embedded media files in archive file formats
* Adds global configuration options for compression
* Fixes windows tests
* Fixes Ubuntu/Python 3.6 tests

### DIFF
--- a/ricecooker/utils/zip.py
+++ b/ricecooker/utils/zip.py
@@ -20,7 +20,7 @@ def _read_file(path):
         return f.read()
 
 
-def create_predictable_zip(path, entrypoint=None):
+def create_predictable_zip(path, entrypoint=None, file_converter=None):
     """
     Create a zip file with predictable sort order and metadata so that MD5 will
     stay consistent if zipping the same content twice.
@@ -29,6 +29,7 @@ def create_predictable_zip(path, entrypoint=None):
         entrypoint (str or None): if specified, a relative file path in the zip to serve as the first page to load
     Returns: path (str) to the output zip file
     """
+    extension = "zip"
     # if path is a directory, recursively enumerate all the files under the directory
     if os.path.isdir(path):
         paths = []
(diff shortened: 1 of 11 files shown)


ricecooker/utils/zip.py
```diff
@@ -47,24 +48,32 @@ def reader(x):
...
-            return inputzip.read(x)
+            def reader(x):
+                return inputzip.read(x)
 
-    else:
-        raise Exception("The `path` must either point to a directory or to a zip file.")
+        except Exception:
+            raise Exception(
+                "The `path` must either point to a directory or to a zip archive."
+            )
```
nucleogenesis, line 63: I feel like I've been yelled at by a linter for trying to except a base `Exception` and had to make my own subclass of it to pass - is this due to lack of parity between Ricecooker and other repos on linting or am I misremembering?

tests/media_utils/test_audio.py
```diff
@@ -0,0 +1,101 @@
...
+        os.remove(filename)
+    except FileNotFoundError:
+        pass
+    assert not os.path.exists(filename)
+
+
+class TempFile(object):
+    """
+    tempfile.NamedTemporaryFile deletes the file as soon as the filehandle is closed.
+    This is OK on unix but on Windows the file can't be used by other commands
```
nucleogenesis, line 84: Is this problem resolved primarily by passing `delete=True` to the `NamedTemporaryFile` and then the context manager business just ensures no matter what w/ the `atexit` registration that the files are always deleted?

nucleogenesis (approved):
Looks good just a few questions overall.

I wonder if that context manager `TempFile` would be helpful to use in any of the production code where we use `NamedTemporaryFile`? Maybe there is some risk of problems on Windows that this being exclusively used in the test suite obfuscates?

I suppose that if there haven't been reports of such issues from our Windows user that it probably isn't the case, but next time I mess with any kind of temporary files I'll probably come borrow that context manager class to simplify and be sure of the clean up.

ricecooker/utils/zip.py
rtibbles: The only thing I can think of that a linter might shout at you for is a bare `except` clause without an `Exception` class or subclass specified.

tests/media_utils/test_audio.py
rtibbles: So, it's to do with accessing the file from a subprocess (using e.g. ffmpeg), this basically lets us treat it like a regular context manager for a temporary file, but still have the deletion.

Note that I copied this from the `test_video.py` and did not come up with this myself, I think...

PR #412, opened 2023-03-25
Add StudioContentNode class to allow scripted curation of existing content from Studio

jamalex:
In order to support the ability to create a new channel that includes content nodes that already exist on Kolibri Studio, this PR introduces a new node class called `StudioContentNode`, which is provided with a reference to the IDs of content nodes that already exist on Studio, along with potential overrides of existing metadata such as the title, tags, or thumbnail. Studio then clones the existing node from its source channel and puts it into the appropriate spot in this newly cheffed channel.

This PR depends on changes @rtibbles shall shortly open a PR for on https://github.com/learningequality/studio.

### DIFF
--- a/examples/remotecontent/sushichef.py
+++ b/examples/remotecontent/sushichef.py
@@ -15,6 +15,7 @@ curates content from another channel already on Studio into a new channel.
 
 SOURCE_DOMAIN = "testdomain.org"  ## change me!
 
+# global dict to retain state between the two chef runs
 original_channel_data = {
     "channel_id": None,
     "doc_node_id": None,
--- a/examples/remotecontent/sushichef.py
+++ b/examples/remotecontent/sushichef.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python
+from ricecooker.chefs import SushiChef
+from ricecooker.classes.files import AudioFile
+from ricecooker.classes.files import DocumentFile
+from ricecooker.classes.licenses import get_license
+from ricecooker.classes.nodes import AudioNode
+from ricecooker.classes.nodes import DocumentNode
+from ricecooker.classes.nodes import RemoteContentNode
+from ricecooker.classes.nodes import TopicNode
+
+"""
+This example shows how to use the RemoteContentNode to create a channel that
+curates content from another channel already on Studio into a new channel.
+"""
+
+SOURCE_DOMAIN = "testdomain.org"  ## change me!
+
+original_channel_data = {
+    "channel_id": None,
+    "doc_node_id": None,
+    "audio_node_id": None,
+}
+
+
+class OriginalChannelChef(SushiChef):
+    channel_info = {
+        "CHANNEL_TITLE": "Original channel",
+        "CHANNEL_SOURCE_DOMAIN": SOURCE_DOMAIN,
+        "CHANNEL_SOURCE_ID": "originalchannel",
+        "CHANNEL_LANGUAGE": "en",
+    }
+
+    def construct_channel(self, **kwargs):
+        channel = self.get_channel(**kwargs)
+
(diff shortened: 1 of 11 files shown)


ricecooker/classes/nodes.py
```diff
@@ -615,6 +619,68 @@ def validate(self):
             assert len(tag) <= 30, (
                 "ERROR: tag " + tag + " is too long. Tags should be 30 chars or less."
             )
+
```
jamalex, line 622: This was just moved up from a subclass to ensure it gets run for all node types.

jamalex: (same with the "role" stuff above)

examples/remotecontent/sushichef.py
```diff
@@ -0,0 +1,126 @@
...
+from ricecooker.classes.nodes import TopicNode
+
+"""
+This example shows how to use the RemoteContentNode to create a channel that
+curates content from another channel already on Studio into a new channel.
+"""
+
+SOURCE_DOMAIN = "testdomain.org"  ## change me!
+
+original_channel_data = {
```
nucleogenesis, line 18: As I read through this it was odd that this null-filled dict was being referenced in a class definition until I got to the bottom and saw how the values were being set before the constructor call. 

I think that this would read a bit more clearly if the values for this were passed as kwargs to the CuratedChannelChef constructor.

nucleogenesis (approved):
Code looks good to me and the examples make it clear how it can be used - just had non-blocking thought on the example code.

One thought was that it might be a bit more consistent and clear to use `Studio` in place of `Remote` in the naming scheme since that's where we're getting the node from.

examples/remotecontent/sushichef.py
jamalex: The reason for this was that we need somewhere "global" to store the computed IDs that are generated during the creation of the original channel -- which would be messy to pull out later through introspection of that channel.

In terms of how they then get passed into the curated chef, I agree that arguments would be nicer. But currently [the kwargs passed into a chef constructor](https://github.com/jamalex/ricecooker/blob/d464ee9863f5807af28058be7766be17811cd7dd/ricecooker/chefs.py#L53) are then just ignored completely and not stored anywhere. There are lots of things around these ricecooker idioms I'd love to rearchitect, but I was picking my battles for this PR.

jamalex: (For now I'll add a comment here to make this clearer, though)

jamalex:
> One thought was that it might be a bit more consistent and clear to use `Studio` in place of `Remote` in the naming scheme since that's where we're getting the node from.

Yeah, this was @rtibbles' recommendation to match the naming pattern of `RemoteFile` that's already in there (used to avoid needing to reupload the same file later when it's already on Studio). I agree that a `Studio` prefix would be more explicit (since it doesn't allow an arbitrary remote URL etc). @rtibbles, what do you think of a rename across both? (with perhaps a deprecated alias for `RemoteFile` for backcompat?)

rtibbles:
This change seems fine, but the backcompat alias would be helpful.

Then in 3 years' time when we allow ricecooker to run against Kolibri we can rename it to `Remote` again and leave the `Studio` alias for back compat :)

jamalex:
- Added comment to clarify global dict purpose
- Changed names of this and `RemoteFile`, with backcompat aliases
- Fixed `IncompleteRead` tests based on [this](https://github.com/DataDog/dd-trace-py/blob/1.x/.github/workflows/test_frameworks.yml#L348) by downgrading `urllib3` version. Would it be preferable to run with that version only in the Github Actions context, or have the test/production envs be consistent?

rtibbles (approved):

PR #415, opened 2023-04-26
Added encoding on reading json file

AtKristijan:
We have encoding on write tree to json, but we didn't have encoding utf8 on read_tree_from_json


rtibbles (approved):
Encodings matter! This looks good, merging.

PR #459, opened 2024-01-12
Drop Python 3.7 support.

rtibbles:


rtibbles:
Merging - will fix the docs build separately.

PR #468, opened 2024-02-19
Initial integration of QTI and IMSCP import 

### LINKED ISSUE #332: Decompose IMS Content Packages (SCORM/IMSCP) into individual Kolibri resources
## Overview

Migrate the `imscp` code into ricecooker.

Decompose an IMS Content Package (including SCORM) into individual Kolibri resources, driven by its manifest.

This reverses the original "no dependency zips" decision; shared-library dependency zips return as a follow-up.

**Complexity:** High
**Target branch:** main

## Context

- The `imscp` code lives in a separate repo (`learningequality/imscp`) and should be native to ricecooker.
- IMSCP is the packaging layer under SCORM.
- Its manifest (`imsmanifest.xml`) is the standardized, tool-independent structure.
- One opaque webcontent zip backs many nodes, so Kolibri's per-resource import pulls the whole package when a single node is selected.
- Wrapping plain media in an HTML5 sandbox strips native affordances — a video should play as a VideoNode, not a `<video>` in a sandbox.
- The existing reference machinery locates and rewrites references only within a single archive.

## The decomposition ladder

Per manifest resource, take the highest rung it safely qualifies for:

1. Native media node (video/audio/pdf/image) — a single media file in a wrapper.
2. KPUB — static article/document HTML.
3. HTML5 zip served via zipcontent — genuinely interactive content.
4. QTI assessment item — assessment content.

Classification runs on one axis: does the resource's meaning survive stripping its JS/CSS?

- Discount SCORM API boilerplate (`pipwerks`, `LMSInitialize`, `SetValue`) before judging interactivity.
- Bias conservative — promote only on high confidence.
- Never promote a resource carrying assessment, score, or tracking semantics.

## The crux

Decomposition must rewrite references *across* resources: a promoted node referenced by a sibling, or a shared library in a dependency zip.

The rewrite emits two new target forms:

- native-node references
- zipcontent-relative dependency paths

## Deferred to follow-ups

- Per-tool deep parsers (Storyline, Captivate, Rise, iSpring, Adapt content models).
- Interactivity classification beyond the sanitization-survival test.
- Automatic dependency-zip vendoring for shared HTML5 libraries.

## Acceptance Criteria

- [ ] An IMSCP/SCORM package imports natively, with the `imscp` code migrated into ricecooker.
- [ ] The manifest drives the node/topic tree.
- [ ] A single wrapped media file becomes a native media node.
- [ ] A static-article resource becomes a KPUB.
- [ ] An interactive resource becomes an HTML5 zip served via zipcontent.
- [ ] A resource carrying assessment semantics is never promoted to media or KPUB.
- [ ] References between decomposed resources resolve correctly in Kolibri.
- [ ] Each decomposed node is backed by its own files, not one shared package zip.

### LINKED ISSUE #337: Handle QTI exercises
Read metadata from and handle QTI exercises to upload

rtibbles:
* Does an initial integration of IMSCP import into ricecooker.
* Adds some simple utilities for mapping IMSCP/SCORM metadata fields to our metadata
* Allows proper mapping of embedded webcontent files to the entry option
* Writes tests to cover this behaviour

Fixes #332 
Fixes #337

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -288,6 +288,26 @@ class HTMLZipFile(DownloadFile):
     default_preset = format_presets.HTML5_ZIP
 
 
+class IMSCPZipFile(DownloadFile):
+    default_ext = file_formats.HTML5
+    allowed_formats = {file_formats.HTML5}
+    is_primary = True
+
+    def get_preset(self):
+        return self.preset or format_presets.IMSCP_ZIP
+
+
+class QTIZipFile(DownloadFile):
+    """Placeholder for QTI zip file support. QTI pipeline handlers are not yet implemented."""
+
+    default_ext = file_formats.HTML5
+    allowed_formats = {file_formats.HTML5}
+    is_primary = True
+
+    def get_preset(self):
+        return self.preset or format_presets.QTI_ZIP
+
+
 class H5PFile(DownloadFile):
     default_ext = file_formats.H5P
     allowed_formats = {file_formats.H5P}
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -298,7 +298,11 @@ class IMSCPZipFile(DownloadFile):
 
 
 class QTIZipFile(DownloadFile):
-    """Placeholder for QTI zip file support. QTI pipeline handlers are not yet implemented."""
+    """File class for QTI (Question and Test Interoperability) zip packages.
+
+    Used for exercise-kind leaf nodes within IMSCP packages that contain
+    QTI assessment resources (resource type prefix ``imsqti_``).
+    """
 
     default_ext = file_formats.HTML5
     allowed_formats = {file_formats.HTML5}
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -1,4 +1,5 @@
 # Node models to represent channel's tree
+import copy
 import json
 import re
 import uuid
@@ -795,13 +796,27 @@ class TreeNode(Node):
         metadata = self.parent.gather_ancestor_metadata()
         return self.get_metadata_dict(metadata)
 
-    def _build_children_from_metadata(self, children_list, parent_files=None, license=None):
-        """Build child TopicNode/ContentNode tree from nested ContentNodeMetadata dicts.
+    def _build_children_from_metadata(
+        self, children_list, parent_files=None, license=None
+    ):
+        """Build child TopicNode/ContentNode tree from nested ContentNodeMetadata.
 
         Called when the pipeline produces nested metadata (e.g. IMSCP manifests).
 
+        File copy semantics:
+          - Leaf children with a ``file_preset`` receive shallow copies
+            (``copy.copy``) of the parent files, with the copy's preset
+            overridden to the child's ``file_preset``.
+          - Leaf children without a ``file_preset`` receive no files.
+          - TopicNode children never receive files.
+
+        Note: Leaf children are created as ContentNode which requires a license.
+        The license defaults to self.license, so this method is primarily suited
+        for calls from ContentNode (which has a license attribute).  Calling from
+        TopicNode requires passing the license argument explicitly.
+
         Args:
-            children_list: list of child metadata dicts
+            children_list: list of ContentNodeMetadata objects
             parent_files: files to propagate to leaf children (defaults to self.files)
             license: license to propagate to leaf ContentNode children (defaults to self.license)
         """
@@ -809,29 +824,33 @@ class TreeNode(Node):
             parent_files = list(self.files)
         license = license or getattr(self, "license", None)
 
-        for child_data in children_list:
-            source_id = child_data.get("source_id", child_data.get("title", "unknown"))
-            title = child_data.get("title", "Untitled")
+        for child_meta in children_list:
+            source_id = child_meta.source_id or child_meta.title or "unknown"
+            title = child_meta.title or "Untitled"
 
-            nested_children = child_data.get("children")
-
-            if nested_children:
+            if child_meta.children:
                 # This is a topic node
                 node = TopicNode(
                     source_id,
                     title,
                 )
                 node._build_children_from_metadata(
-                    nested_children, parent_files=parent_files, license=license
+                    child_meta.children, parent_files=parent_files, license=license
                 )
             else:
-                # This is a leaf content node — filter files by preset if specified
-                file_presets = child_data.get("file_presets", [])
-                if file_presets:
-                    files = [f for f in parent_files if f.get_preset() in file_presets]
+                # This is a leaf content node — copy files with child's preset.
+                # Shallow copy is safe here: File attributes are all simple
+                # types (str, int, None), so no mutable state is shared.
+                file_preset = child_meta.file_preset
+                if file_preset:
+                    files = []
+                    for f in parent_files:
+                        f_copy = copy.copy(f)
+                        f_copy.preset = file_preset
+                        files.append(f_copy)
                 else:
                     files = []
-                extra_fields = child_data.get("extra_fields", {})
+                extra_fields = child_meta.extra_fields or {}
                 node = ContentNode(
                     source_id,
                     title,
@@ -839,7 +858,7 @@ class TreeNode(Node):
                     files=list(files),
                     extra_fields=extra_fields,
                 )
-                node.kind = child_data.get("kind", content_kinds.HTML5)
+                node.kind = child_meta.kind or content_kinds.HTML5
                 node._files_processed = True
 
             # Apply additional metadata fields
@@ -852,8 +871,8 @@ class TreeNode(Node):
                 "language",
             ]
             for field in metadata_fields:
-                if field in child_data:
-                    val = child_data[field]
+                val = getattr(child_meta, field, None)
+                if val is not None:
                     existing = getattr(node, field, None)
                     if isinstance(existing, list) and isinstance(val, list):
                         setattr(node, field, existing + val)
@@ -979,6 +998,37 @@ class ContentNode(TreeNode):
             self._validate_uri()
         return super(ContentNode, self)._validate()
 
+    _FILE_INIT_KEYS = {
+        "preset",
+        "language",
+        "default_ext",
+        "source_url",
+        "duration",
+        "original_filename",
+        "filename",
+    }
+
+    # Fields from ContentNodeMetadata that _process_uri may apply to the node.
+    # Identity/legal fields (source_id, license, license_description,
+    # copyright_holder) are intentionally excluded to prevent pipeline
+    # metadata from overwriting values set in the constructor.
+    _METADATA_APPLY_FIELDS = (
+        "title",
+        "description",
+        "thumbnail",
+        "author",
+        "aggregator",
+        "provider",
+        "language",
+        "grade_levels",
+        "categories",
+        "resource_types",
+        "learning_activities",
+        "accessibility_labels",
+        "learner_needs",
+        "role",
+    )
+
     def _process_uri(self):
         try:
             file_metadata_list = self.pipeline.execute(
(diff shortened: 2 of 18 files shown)


ricecooker/utils/SCORM_metadata.py
```diff
@@ -0,0 +1,173 @@
...
+        return [needs.FOR_BEGINNERS]
+
+    return []
+
+
+def update_node_from_metadata(node, metadata_dict):
+    # Update the node with the general metadata
+    node.description = metadata_dict.get("description") or node.description
+    if metadata_dict.get("language"):
+        node.set_language(metadata_dict.get("language"))
```
jredrejo, line 159: I had two problems ithe these few lines of code:
language was en-US
metadata.dict was this one:
```
(Pdb) metadata_dict
{'title': 'Work with computers', 'description': 'This learning path will introduce you to the different parts and types of the computer and their functions. You will also learn the difference between operating systems and applications and their functions. Peripherals and portable storage devices will be discussed as well.', 'language': 'en-US', 'keyword': 'Default'}
```

It worked with these changes (I ignore if the `keyword` change will work with other scorm files)
```
    if metadata_dict.get("language"):
        node.set_language(metadata_dict.get("language").lower().split("-")[0])
    keyword = metadata_dict.get("keyword")
    if keyword:
        node.tags = node.tags + [keyword,]
```

rtibbles: Updated - hopefully has a similar effect, add some additional validation to make sure it's a language we know.

ricecooker/classes/files.py
```diff
@@ -768,6 +774,210 @@ def process_file(self):
...
+    return value
+
+
+class IMSCPZipFile(DownloadFile):
+    default_ext = file_formats.HTML5
+    allowed_formats = [file_formats.HTML5]
+    is_primary = True
+
+    def get_preset(self):
+        return self.preset or format_presets.IMSCP_ZIP
```
jredrejo, line 807: This is producing a server error in Studio https://learningequality.sentry.io/issues/4997106816/?project=1252819

jredrejo:
I've been playing with the latest code from this PR:

- Ricecooker works
- Uploading to Studio unstable works
- The channel appears correctly (preview doesn't work, but that's not a ricecooker problem)  https://unstable.studio.learningequality.org/es-es/channels/6877a75b029e5259a02c93046ee6dd80 ,   token: kokom-fitih
- Using kolibri v0.16.1, I can see the channel, the structure is a bit weird, with more levels than expected: 1 single scorm file has up to four levels to access to the resource:
image
- After importing it, it does not work correctly, this is the error message:
image

jredrejo (changes requested):
Left a detailed comment with the two issues found

rtibblesbot (changes requested):
Solid initial integration of IMSCP import into ricecooker — the manifest parsing, SCORM metadata mapping, and pipeline handler structure are well-designed. Two blocking issues around license propagation and dynamic child processing in nested structures need fixing before merge.

CI passing. All tests pass across Python 3.9-3.13 on all platforms.

**Blocking:**
- License not propagated through nested topics — leaf `ContentNode`s under `TopicNode`s get `license=None` (see inline)
- `process_files` dynamic children loop doesn't recurse into `TopicNode` children's descendants (see inline)
- `assert` used for runtime validation of untrusted XML input in `imscp.py` — disabled under `python -O` (see inline)

**Suggestions:**
- `QTIZipFile` added to `files.py` but no corresponding QTI handler or tests — consider deferring to a follow-up PR or adding a comment (see inline)
- `_extract_lom_text` can return mixed types (string, list, or None) which could surprise callers (see inline)

rtibblesbot (commented):
Solid initial integration of IMSCP and QTI import into ricecooker. The architecture is well-considered — standalone manifest parsing, metadata mapping utilities, pipeline handler integration, and nested child node building all fit cleanly into the existing codebase patterns.

CI passing (linting). Test coverage is thorough across all new modules.

- **suggestion** (1): Consider handler ordering for edge cases (convert.py)
- **suggestion** (2): Potential `AttributeError` when `TreeNode._build_children_from_metadata` is called on a `TopicNode` with no `files` attribute default (nodes.py)
- **suggestion** (3): `_apply_content_node_metadata` unconditionally overwrites list fields from pipeline (nodes.py)
- **nitpick** (1): Metadata field application list is duplicated between two methods (nodes.py)
- **nitpick** (2): `_recursive_update` could be a `dict` in-place update note (context.py)

ricecooker/classes/nodes.py
```diff
@@ -767,6 +768,91 @@ def gather_ancestor_metadata(self):
...
+        TopicNode requires passing the license argument explicitly.
+
+        Args:
+            children_list: list of ContentNodeMetadata objects
+            parent_files: files to propagate to leaf children (defaults to self.files)
+            license: license to propagate to leaf ContentNode children (defaults to self.license)
+        """
+        if parent_files is None:
+            parent_files = list(self.files)
+        license = license or getattr(self, "license", None)
```
rtibblesbot, line 797: **suggestion:** `license = license or getattr(self, "license", None)` — if `self.license` is a falsy value (e.g. an empty string was somehow set), this would fall through to `None`. In practice this is likely fine since `ContentNode.__init__` requires a license object, but worth noting that `getattr` with a default of `None` combined with `or` means the fallback is only used when the attribute is missing *or* falsy, not just missing.

ricecooker/classes/nodes.py
```diff
@@ -767,6 +768,91 @@ def gather_ancestor_metadata(self):
...
+                    title,
+                    license,
+                    files=list(files),
+                    extra_fields=extra_fields,
+                )
+                node.kind = child_meta.kind or content_kinds.HTML5
+                node._files_processed = True
+
+            # Apply additional metadata fields
+            metadata_fields = [
```
rtibblesbot, line 837: **nitpick:** This `metadata_fields` list overlaps significantly with `_METADATA_APPLY_FIELDS` (defined on `ContentNode` at line 998). The two lists serve the same purpose (applying metadata to a node) but differ slightly (this one includes `tags` but omits `thumbnail`, `author`, `aggregator`, `provider`, `role`, etc.). Consider extracting a shared constant, or documenting why the child-building path intentionally uses a narrower set of fields.

ricecooker/classes/nodes.py
```diff
@@ -903,29 +1020,83 @@ def _process_uri(self):
...
+            k: v for k, v in metadata_dict.items() if k in self._FILE_INIT_KEYS
+        }
+        self.add_file(File(**file_kwargs))
+
+    def _apply_content_node_metadata(self, content_node_metadata):
+        """Apply ContentNodeMetadata fields to this node and build children.
+
+        See _METADATA_APPLY_FIELDS for the allowlist and exclusion rationale.
+        """
+        for field_name in self._METADATA_APPLY_FIELDS:
```
rtibblesbot, line 1054: **suggestion:** `_apply_content_node_metadata` unconditionally overwrites list fields like `learning_activities`, `resource_types`, `grade_levels`, etc. via `setattr`. If the node had pre-existing values set in the constructor (e.g. by the sushi chef), the pipeline metadata would silently replace them rather than merge. Contrast with the `tags` handling two lines below (line 1061) which correctly appends. This may be intentional (pipeline metadata should take precedence for these fields) but it's worth a comment documenting the design choice, or making the merge-vs-replace behavior consistent.

ricecooker/utils/pipeline/extract_metadata.py
```diff
@@ -64,6 +72,79 @@ class PDFMetadataExtractor(MetadataExtractor):
...
+        # IMSCP for: pure webcontent, mixed (QTI + webcontent), or
+        # unknown resource types (safe default for any IMS package)
+        return format_presets.IMSCP_ZIP
+
+    def _build_children_metadata(self, items):
+        """Recursively build ContentNodeMetadata list from parsed item dicts."""
+        children = []
+        for item in items:
+            fields = metadata_dict_to_content_node_fields(item.get("metadata", {}))
+            fields["title"] = item.get("title", fields.get("title"))
```
rtibblesbot, line 106: **suggestion:** `fields["title"] = item.get("title", fields.get("title"))` — if the item has no `"title"` key and `metadata_dict_to_content_node_fields` also produced no title, `fields["title"]` will be set to `None`. Downstream in `_build_children_from_metadata` (nodes.py line 801), the fallback `child_meta.title or "Untitled"` handles this, but `_content_node_metadata_from_dict` will store `None` as the title. This is fine but slightly surprising — a missing title propagates as `None` through the metadata layer and only gets defaulted at the node-building layer.

ricecooker/utils/imscp.py
```diff
@@ -0,0 +1,474 @@
...
+            logger.warning("Dangling identifierref: %s", item["identifierref"])
+            return
+
+        for key, value in resource_elem.items():
+            key_stripped = re.sub("^{.*}", "", key)
+            if key_stripped not in item:
+                item[key_stripped] = value
+
+        resource_type = resource_elem.get("type", "")
+        if resource_type == "webcontent" or is_qti_resource(resource_type):
```
rtibblesbot, line 310: **suggestion:** `if resource_type == "webcontent" or is_qti_resource(resource_type):` — any other resource types (e.g. `"associatedcontent/imscc_xmlv1p1/learning-application-resource"`) won't get `files` populated. This means resources of unrecognized types are silently dropped. Consider logging a warning for unrecognized resource types that have file references, similar to how dangling `identifierref` is logged on line 301.

ricecooker/utils/imscp.py
```diff
@@ -0,0 +1,474 @@
...
+    try:
+        with zf.open("imsmanifest.xml") as manifest_file:
+            return etree.parse(manifest_file, _SAFE_PARSER).getroot()
+    except (etree.XMLSyntaxError, OSError):
+        pass
+
+    # Handle XML files that are marked as UTF-8 but have non-UTF-8 chars.
+    # Detect the real encoding with chardet and re-encode as UTF-8.
+    try:
+        f = zf.open("imsmanifest.xml", "r")
```
rtibblesbot, line 206: **nitpick:** The `chardet` fallback opens the file in a non-context-manager style (`f = zf.open(...); data = f.read(); f.close()`). Using a `with` statement would be marginally cleaner:
```python
with zf.open("imsmanifest.xml", "r") as f:
    data = f.read()
```

ricecooker/utils/pipeline/extract_metadata.py
```diff
@@ -65,6 +73,79 @@ class PDFMetadataExtractor(MetadataExtractor):
...
+        organizations/leaf items).
+        """
+        has_qti = has_qti_items(manifest)
+        has_webcontent = has_webcontent_items(manifest)
+        if has_qti and not has_webcontent:
+            # Pure QTI package (assessments only)
+            return format_presets.QTI_ZIP
+        # IMSCP for: pure webcontent, mixed (QTI + webcontent), or
+        # unknown resource types (safe default for any IMS package)
+        return format_presets.IMSCP_ZIP
```
rtibblesbot, line 100: **praise:** Defaulting to `IMSCP_ZIP` for mixed packages and unknown resource types is the right call — it handles the open-ended case (e.g. proprietary LMS resource types, future IMS resource types) gracefully without silently dropping content or crashing. The comment spelling out the covered cases makes the intent clear.

ricecooker/classes/nodes.py
```diff
@@ -795,6 +796,91 @@ def gather_ancestor_metadata(self):
...
+        TopicNode requires passing the license argument explicitly.
+
+        Args:
+            children_list: list of ContentNodeMetadata objects
+            parent_files: files to propagate to leaf children (defaults to self.files)
+            license: license to propagate to leaf ContentNode children (defaults to self.license)
+        """
+        if parent_files is None:
+            parent_files = list(self.files)
+        license = license or getattr(self, "license", None)
```
rtibblesbot, line 825: **suggestion (re-raise):** `license = license or getattr(self, "license", None)` — if `self.license` is somehow a falsy value (empty string, zero), the `or` would fall through to `None`, which would then fail validation when the leaf `ContentNode` is created. The docstring correctly notes this method is suited for `ContentNode` callers, but a guard like `if license is None: license = getattr(self, "license", None)` would make the None-only fallback intent explicit and immune to falsy values.

ricecooker/classes/nodes.py
```diff
@@ -795,6 +796,91 @@ def gather_ancestor_metadata(self):
...
+                    title,
+                    license,
+                    files=list(files),
+                    extra_fields=extra_fields,
+                )
+                node.kind = child_meta.kind or content_kinds.HTML5
+                node._files_processed = True
+
+            # Apply additional metadata fields
+            metadata_fields = [
```
rtibblesbot, line 865: **nitpick (re-raise):** `metadata_fields` here overlaps significantly with `_METADATA_APPLY_FIELDS` (line 1015) — both serve the purpose of "apply these metadata fields to a node" but differ slightly (this one includes `tags` and `learner_needs`; `_METADATA_APPLY_FIELDS` adds `title`, `thumbnail`, `author`, etc.). Extracting a shared constant, or at least a comment explaining why these two lists differ, would reduce the maintenance surface.

ricecooker/classes/nodes.py
```diff
@@ -920,29 +1037,83 @@ def _process_uri(self):
...
+
+    def _apply_content_node_metadata(self, content_node_metadata):
+        """Apply ContentNodeMetadata fields to this node and build children.
+
+        See _METADATA_APPLY_FIELDS for the allowlist and exclusion rationale.
+        """
+        for field_name in self._METADATA_APPLY_FIELDS:
+            value = getattr(content_node_metadata, field_name, None)
+            if value is not None:
+                setattr(self, field_name, value)
```
rtibblesbot, line 1074: **suggestion (re-raise):** `setattr(self, field_name, value)` unconditionally replaces list fields like `learning_activities`, `resource_types`, etc. when the pipeline provides them. Contrast with `_build_children_from_metadata` (line 877: `existing + val`) which merges lists. If a sushi chef pre-populates these fields in the constructor and the pipeline also provides values, the chef's values are silently dropped here. Either document that pipeline metadata takes precedence over constructor values for these fields, or apply the same merge logic used in `_build_children_from_metadata`. (Note: `tags` is correctly merged two lines below.)

ricecooker/utils/pipeline/extract_metadata.py
```diff
@@ -65,6 +73,79 @@ class PDFMetadataExtractor(MetadataExtractor):
...
+
+    def _build_children_metadata(self, items):
+        """Recursively build ContentNodeMetadata list from parsed item dicts."""
+        children = []
+        for item in items:
+            fields = metadata_dict_to_content_node_fields(item.get("metadata", {}))
+            fields["title"] = item.get("title", fields.get("title"))
```
rtibblesbot, line 107: **suggestion (re-raise):** `fields["title"] = item.get("title", fields.get("title"))` — if neither `item` nor the LOM metadata produced a title, this stores `None` as the title key. Downstream in `_build_children_from_metadata` the fallback `child_meta.title or "Untitled"` handles it, but it means `ContentNodeMetadata.title` can be explicitly `None` (from a key that was set to `None`) rather than absent. Not broken, but slightly surprising — worth a comment or a guard like `or fields.pop("title", None)` to keep None out of the dict.

ricecooker/utils/imscp.py
rtibblesbot, line 310: **suggestion (re-raise):** Resources whose `type` is neither `webcontent` nor a `imsqti_*` prefix (e.g. `associatedcontent/imscc_xmlv1p1/learning-application-resource`) fall through with no `files` key set. This silently drops their file references. A `logger.warning` for unrecognized types that have child file elements would help chefs diagnose unexpected empty packages, similar to the dangling-identifierref warning on line 301.

ricecooker/utils/imscp.py
rtibblesbot, line 206: **nitpick (re-raise):** The chardet fallback uses `f = zf.open(...); data = f.read(); f.close()` rather than a `with` block. If `f.read()` raises, `f.close()` is skipped. A `with zf.open("imsmanifest.xml", "r") as f:` is the safe pattern here.

rtibblesbot (changes requested):
### Prior findings

5 prior findings from round 1 all resolved (3 blocking, 2 suggestion — resolved in 1b191bc). Well done on the thorough resolution.

0/6 second-round findings resolved. All 6 re-raised below (no code changes since round 2).

---

CI now failing — `ruff-format` reformatted 5 files when run locally: `ricecooker/utils/SCORM_metadata.py`, `ricecooker/utils/imscp.py`, `tests/helpers.py`, `tests/test_imscp.py`, `tests/test_imscp_pipeline.py`. Round 2 noted "CI passing (linting)" — this appears to be a formatting regression, likely from a prek/ruff version bump on main since February. Fix: run `uvx prek run --all-files` and commit the reformatted files.

**Blocking:**
- ruff-format failing on 5 files — run `uvx prek run --all-files` and commit

**Suggestions and nitpicks** are all re-raises from round 2 (UNADDRESSED — no replies, no code changes since then):
- License fallback has a false-falsy edge case (suggestion, nodes.py)
- `_apply_content_node_metadata` overwrites list fields rather than merging (suggestion, nodes.py)
- `fields["title"]` can be stored as `None` (suggestion, extract_metadata.py)
- Unrecognized resource types silently drop their files (suggestion, imscp.py)
- `metadata_fields` duplicates a subset of `_METADATA_APPLY_FIELDS` (nitpick, nodes.py)
- chardet fallback opens file without a `with` statement (nitpick, imscp.py)

---

Compared the current PR state against findings from a prior review:
- Retrieved prior bot reviews deterministically via the GitHub API
- Classified each prior finding as RESOLVED, UNADDRESSED, ACKNOWLEDGED, or CONTESTED
- Only raised NEW findings for newly introduced code
- Reviewed the pull request diff checking for correctness, design, architecture, testing, completeness, and adherence to DRY/SRP principles
- Checked CI status and linked issue acceptance criteria

</details>

rtibbles:
Superseded by #721

PR #469, opened 2024-02-21
Use the right object to get the error information

jredrejo:
In the current code, when a POST to https://api.studio.learningequality.org/api/file/upload_url does not return a 200 code, the exception is  using the wrong object to build the error message


rtibbles (approved):
Woops, pretty sure this was my fault! Thanks @jredrejo

PR #477, opened 2024-03-04
Download css background images

jredrejo:
When downloading a page using ArchiveDownloader, background images references in css styles where not being fetched.
This PR adds support for it.

### DIFF
--- a/ricecooker/utils/downloader.py
+++ b/ricecooker/utils/downloader.py
@@ -361,7 +361,7 @@ def download_static_assets(  # noqa: C901
                 continue
 
             if "background-image" in selector:
-                url_match = re.search(r'url\("([^"]+)"\)',node['style'])
+                url_match = re.search(r'url\("([^"]+)"\)', node["style"])
                 if url_match:
                     background_image_url = url_match.group(1)
                     url = urljoin(base_url, background_image_url)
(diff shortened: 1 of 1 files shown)


ricecooker/utils/downloader.py
```diff
@@ -360,7 +360,15 @@ def download_assets(
             if node[attr].startswith("data:"):
                 continue
 
-            url = urljoin(base_url, node[attr])
+            if "background-image" in selector:
+                url_match = re.search(r'url\("([^"]+)"\)', node["style"])
```
rtibbles, line 364: I don't think this will catch all cases of the CSS `url` function, see here for a comprehensive example in JS regex: https://github.com/learningequality/kolibri/blob/develop/packages/kolibri-zip/src/fileUtils.js#L33

rtibbles (changes requested):
Looks good, I think the url regex can be made more general though.

Adding some test cases to cover the different permutations would be helpful, see the parallel JS tests I made for the referenced function here: https://github.com/learningequality/kolibri/blob/develop/packages/kolibri-zip/test/fileUtils.spec.js#L20

ricecooker/utils/downloader.py
jredrejo: I'm not going to argue with any regex string. I have adapted the example in js to python and changed this code after I tested it works.

rtibbles (approved):

PR #482, opened 2024-03-11
docs: update link to api token

ThEditor:
The current link to the API token page is outdated. I've updated the link to directly point to the "settings" page which now holds the api token.


rtibbles (approved):

rtibbles:
Thank you!

PR #484, opened 2024-03-13
Update content disposition parsing and add tests.

rtibbles:
Fixes bug in Content-Disposition parsing whereby the trailing `"` would be left in the file extension.

Adds tests to robustly guarantee how we parse the header.


jredrejo (approved):
Tested without any issue, and code LGTM

PR #496, opened 2024-04-29
Set default learning_activity based on node kind in TreeNode

### LINKED ISSUE #451: Set default value for `TreeNode.learning_activities` when not set explicitly
## Observed behavior

A `TreeNode` can be shipped off to Studio without a value in `learning_activities` which has been handled in Kolibri  with [`kolibri/core/content/constants/kind_to_learningactivity.py`
](https://github.com/learningequality/kolibri/blob/release-v0.16.x/kolibri/core/content/constants/kind_to_learningactivity.py) but is not done in Studio. 

## Expected behavior

When a `TreeNode` is created without a value in `learning_activities` we should be setting a default if it is empty. 

## Acceptance criteria

- [ ] Identify the ideal place to define the mapping linked above.
- [ ] Utilize that mapping to ensure that when a `TreeNode` is created that we set the default value into the `learning_activities` array based on that node's kind. **But only if we aren't given a value for `learning_activities` to begin with.**

nucleogenesis:
## Overview

In the TreeNode constructor, use [kind_activity_map](https://github.com/learningequality/kolibri/blob/release-v0.16.x/kolibri/core/content/constants/kind_to_learningactivity.py) borrowed from Kolibri to map the node's `kind` to a particular learning activity.

If `learning_activities` is provided to the constructor, then that is used. However, if it's not provided or is empty, then we refer to the map to provide a default learning activity.

Closes #451 

## Reviewer guidance

Are the two little tests sufficient? Any possible regressions?

### DIFF
--- a/.github/workflows/pythontest.yml
+++ b/.github/workflows/pythontest.yml
@@ -32,12 +32,12 @@ jobs:
       max-parallel: 5
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest, macos-13]
         python-version: [3.8, 3.9, '3.10']
         include:
         - os: ubuntu-latest
           pippath: ~/.cache/pip
-        - os: macos-latest
+        - os: macos-13
           pippath: ~/Library/Caches/pip
         - os: windows-latest
           pippath: ~\AppData\Local\pip\Cache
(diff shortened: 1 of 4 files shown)


nucleogenesis:
Would this be worthwhile to put into LE Utils instead @rtibbles ?

Or -- maybe as a follow-up to put onto the list of things we'll do when we get around to reworking LE Utils down the road?

rtibbles (approved):

PR #510, opened 2024-07-17
Test under Python 3.11 & 3.12

rodjek:
Build and test against Python 3.11 and 3.12

Tests pass locally
```
  py3.12: OK (56.96=setup[2.67]+cmd[54.29] seconds)
  congratulations :) (57.01 seconds)
```


rtibbles (approved):
Thanks for this @rodjek!

Issue #501, opened 2024-06-04
Allow integration of BloomPub files


rtibbles:
Add a file class for handling BloomPub files.
Make it possible to upload these files as part of the document node class.
```[tasklist]
### Interdependencies
- [ ] https://github.com/learningequality/studio/issues/4583
```

rtibbles:
This was completed in #517

PR #36, opened 2017-01-05
Version-compatible check

jayoshih:
Note: Need to merge/deploy this to avoid errors: https://github.com/fle-internal/content-curation/pull/127

PR #517, opened 2024-08-02
Allow uploading of bloom format files

nikkuAg:
Added the integration for uploading Bloom format files.
#501 
[#36](https://github.com/learningequality/kolibri-ecosystem/issues/36)


rtibbles (approved):
Useful bugfix, and new file type is properly implemented!

Issue #522, opened 2024-09-02
YouTube download is unmocked in tests


jredrejo:
## Observed behavior

Ricecooker tests are failing due to YouTube errors

## Errors and logs
```
WARNING  ERROR: [youtube] zzJLYK893gQ: Sign in to confirm you’re not a bot. This helps protect our community. Learn more
ERROR    ==> [YouTubeTypes.YOUTUBE_VIDEO (test-video)] Failed to extract YouTube info

```

## Expected behavior

ricecooker tests should pass

## User-facing consequences

Chef scripts using YouTube links will need an account to sign in before retrieving the content

PR #527, opened 2024-10-03
Mock yt-dlp web requests

jredrejo:
Use mocks to avoid tests doing requests to youtube that's filtering GitHub requests.
Closes: #522


rtibbles (approved):
Foxes tests, mocks yt dlp.

Issue #483, opened 2024-03-13
No explicit header set for the DOWNLOAD_SESSION


rtibbles:
## Observed behavior
The DOWNLOAD_SESSION that is used to download resources sets no explicit header - this proves to be an issue, for example, when downloading from wikimedia sites, because of their User Agent policy: https://meta.wikimedia.org/wiki/User-Agent_policy

## Expected behavior
Ideally, we would follow the kind of User-Agent that the wikimedia policy spells out - we already retrieve the email for the user whose API token we are running with from Studio, so we should reuse this to set the header.

With that in place, we would then do the following for the User Agent:

```
f"Ricecooker/{ricecooker.__version__} bot ({user_email})"
```

## User-facing consequences
Attempts to scrape without setting these headers may be treated as malicious.

## Steps to reproduce
Attempt to download any file from wikimedia

## Context
Ricecooker develop branch

rtibbles:
Absolutely @nikkuAg - I will assign you, thanks for volunteering!

MisRob:
Hi @nikkuAg, are you still planning to work on this?

manic012:
Hey, is this issue still open? I would like to work on this

adoo100:
Hi @rtibbles if this issue is still unresolved I would love to be a part and work in it :)

akolson:
Hi @adoo100! Thank you! I have assigned it to you! Please feel free to reach out to @rtibbles incase it is not clear what needs to be done while resolving the issue.

Keshav123454:
Does Ricecooker have there seprate development guide??

AllanOXDi:
Hi @Keshav123454  Here is a starting point https://ricecooker.readthedocs.io/en/latest/developer/index.html

MisRob:
Hi @adoo100,  I wanted to mention that [Learning Equality will be closed from December 23 to January 5](https://github.com/learningequality/kolibri/discussions/12956).

Divyanshi750:
Hi, is this issue still open to work? I would like to contribute to it!

AlexVelezLl:
Hey @adoo100, are you still working on this? Or can we reassign?

AlexVelezLl:
Unassigning due to lack of activity in several months. I will assign this to you @Divyanshi750! Please let us know if you have any question :)

Divyanshi750:
I have fixed the issue and also added test cases for the same. Please review @AlexVelezLl  @rtibbles

[...]

PR #536, opened 2024-11-08
No explicit header set for the DOWNLOAD_SESSION #483

manic012:
…ads a resource from a specified URL while including a custom User-Agent header for identification. It constructs the User-Agent string using the application's version and the user's email, sends an authenticated GET request to the URL, and saves the downloaded content to a file if the request is successful. The script also includes an example usage section where the user can specify their email, API token, and download URL.

### DIFF
--- a/ricecooker/commands.py
+++ b/ricecooker/commands.py
@@ -12,6 +12,8 @@ from .classes.nodes import ChannelNode
 from .managers.progress import RestoreManager
 from .managers.progress import Status
 from .managers.tree import ChannelManager
+from . import ricecooker  # Import to access the version
+
 
 # Fix to support Python 2.x.
 # http://stackoverflow.com/questions/954834/how-do-i-use-raw-input-in-python-3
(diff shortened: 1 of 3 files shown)


ricecooker/commands.py
```diff
@@ -94,6 +96,12 @@ def uploadchannel(  # noqa: C901
         username = ""
         token = ""
 
+    # Set User-Agent header for DOWNLOAD_SESSION
+    user_email = authenticate_user(token)[0]  # Get the user's email
+    config.DOWNLOAD_SESSION.headers.update({
+        "User-Agent": f"Ricecooker/{ricecooker.__version__} bot ({user_email})"
```
rtibbles, line 102: Is the `username` not the user_email in this case?

ricecooker/commands.py
```diff
@@ -94,6 +96,12 @@ def uploadchannel(  # noqa: C901
         username = ""
         token = ""
 
+    # Set User-Agent header for DOWNLOAD_SESSION
+    user_email = authenticate_user(token)[0]  # Get the user's email
```
rtibbles, line 100: Not sure why we need to reauthenticate here, we already got the username above.

rtibbles (changes requested):
I think the change in commands.py is all that is needed.

This rather has the appearance of having been automatically generated by a generative AI, as some of the code here refers to non-existent modules.

Please before submitting a pull request, test that your code actually works and doesn't break things.

manic012:
Ok ,sure i will correct them.

PR #553, opened 2024-12-27
Trusted publishing setup

rtibbles:
## Summary
* Sets up github action for trusted publishing to pypi
* Updates minimum Python version check
* Increments version to 0.8.0
* Moves test file to samples so that it doesn't get cleaned up by `make clean`


.github/workflows/python-publish.yml
```diff
@@ -16,21 +10,21 @@ jobs:
...
       with:
-        python-version: 3.7
+        python-version: 3.9
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine
-    - name: Build and publish
-      env:
-        TWINE_USERNAME: __token__
```
marcellamaki, line 32: so, we're still using twine, we're just no longer accessing the credentials here, right? and instead we're using the `gh-action-pypi-publish`

tests/test_files.py
```diff
@@ -557,18 +542,15 @@ def test_convertible_substitles_from_pressurcooker(pressurcooker_test_files):
...
     """
-    # local_path = os.path.join(
-    #     "tests", "testcontent", "downloaded", "testsubtitles_ar.ttml"
-    # )
     local_path = os.path.abspath(
         os.path.join(
             os.path.dirname(__file__),
             "testcontent",
-            "downloaded",
+            "samples",
```
marcellamaki, line 553: is this changed because we've moved the file path to be under /samples so that it's clearer that it's... a sample? but content-wise it's the same?

tests/test_files.py
```diff
@@ -522,21 +522,6 @@ def pressurcooker_test_files():
     return download_fixture_files(PRESSURECOOKER_SUBS_FIXTURES)
 
 
-@pytest.fixture
-def youtube_test_file():
```
marcellamaki, line 526: I don't think I understand why this is no longer needed

marcellamaki (commented):
just a couple of clarifying questions

.github/workflows/python-publish.yml
rtibbles: Yes, I think internally the package does probably use twine, but we don't have to worry about it any more.

tests/test_files.py
rtibbles: Because it was never used! See the deleted function argument below with the same name (pytest fixtures are referred to in function argument by their names). It was unused there, so I deleted it, and then it wasn't used anywhere else.

tests/test_files.py
rtibbles: Yes - it was originally setup to download, but the download kept failing so it was committed to the repo, but never moved to a folder where that should happen (as opposed to a folder that gets cleaned up by our `make clean` command).

marcellamaki (approved):

Issue #220, opened 2019-08-23
Better error message when user runs out of storage


ivanistheone:
### Description

When a user has run out of storage quota, this path will get triggered:
https://github.com/learningequality/studio/blob/develop/contentcuration/contentcuration/views/internal.py#L134

Current state: Ricecooker currently exits:
<img width="1132" alt="Screen Shot 2019-08-23 at 7 59 28 AM" src="https://user-images.githubusercontent.com/163966/63591268-4114a780-c57c-11e9-83a6-47dec8bc9e6d.png">

Expected: handle error and print informative message to user (you have run out of storage)

devlakshay:
Is it solved?

rtibbles:
I don't think it's giving an informative error message still, as the exception is for any rejected request: https://github.com/learningequality/ricecooker/blob/213a3ca5ae47ce54dd72594184d6187020431a7b/ricecooker/managers/tree.py#L191

Noting that this doesn't prevent any of our retry behaviour, which we should do when there is insufficient storage.

We are also using a different endpoint now, so just checking for a status response of 412 might be sufficient (to see where that 412 is coming from, see here https://github.com/learningequality/studio/blob/unstable/contentcuration/contentcuration/viewsets/file.py#L163).


So, a solution here would be:

* In the else block where we just throw an exception (https://github.com/learningequality/ricecooker/blob/213a3ca5ae47ce54dd72594184d6187020431a7b/ricecooker/managers/tree.py#L191) we should check if the response status code is a 412, and raise a specific exception saying the user is out of storage space.
* Make sure this is propagating through the chain of calls of _handle_upload, upload_files, etc. and make sure we flag (maybe through an attribute on the ChannelManager object) that there is insufficient storage, so that we can raise an error in the main thread, and give a meaningful error to the user.

Divyanshi750:
hey, is this issue still open? I would really like to work on it!

Divyanshi750:
I have fixed the issue by adding an insufficient storage exception. Please review @rtibbles

MisRob:
Thanks @Divyanshi750!

Divyanshi750:
Hi @rtibbles @MisRob,

I have closed the previous PR and raised a new one due to some technical difficulties. In this updated PR:

I have removed changes to unrelated files.
I have added the audio-cache file to .gitignore, as requested.
I sincerely apologize for raising a new PR and appreciate your understanding as I am new to the open-source community. Kindly review it when possible.

Thank you!

[...]

PR #562, opened 2025-01-24
Added the insufficient storage exception

Divyanshi750:
## Summary

This pull request solves the problem of getting an informative error if users run out of storage space; it provides custom InsufficientStorageException for uploading where the status code 412 Precondition Failed will be caught, and this will then be handled appropriately in the main application.

Specifically, I've made the following changes:

1. Defined the InsufficientStorageException class that inherits from the base Exception class

2. In the do_file_upload method of the ChannelManager class, added a check for the 412 status code and raised the InsufficientStorageException if encountered

3. In the _handle_upload method, propagated the InsufficientStorageException up the call stack so that it can be caught and handled in the main application

## References

Original issue: [Better error message when user runs out of storage #220]

## Reviewer guidance

To test these changes you can simulate the 412 precondition failed response by mocking the behavior of the file upload endpoint returned by the post() method config.SESSION to check whether the main application is raised and can correctly catch the InsufficientStorageException.


ricecooker/managers/tree.py
```diff
@@ -144,6 +144,10 @@ def get_file_diff(self, files_to_diff):
                 if not exists
             ]
 
+class InsufficientStorageException(Exception):
```
rtibbles, line 147: These seem like reasonable changes. Please remove changes to the other files.

If you would like to add the `audio_cache.sqlite` file to the `.gitignore` that would be much appreciated.

rtibbles (changes requested):
A little clean up required, but this looks like a reasonable fix.

PR #563, opened 2025-01-24
Headerset

the-ivii:
## Summary

This pull request enhances the User-Agent header generation mechanism in the Ricecooker download utility. The main changes made are:

- Introduced a new `configure_download_session()` function to  generate User-Agent headers
- Modified the `make_request()` function to support optional user email identification

## References

Fixes: #483 
## Reviewer guidance

Added test cases to:
   - Generate requests with various scenarios:
     * No email provided
     * Custom email supplied
     * Different Ricecooker versions


ricecooker/managers/tree.py
```diff
@@ -163,41 +168,45 @@ def do_file_upload(self, filename):
...
-                    .decode()
-                    .strip()
-                )
-                headers = {"Content-Type": content_type, "Content-MD5": b64checksum}
-                response = config.SESSION.put(
-                    upload_url, headers=headers, data=file_obj
-                )
-                if response.status_code == 200:
-                    return
+        url_response = config.SESSION.post(config.get_upload_url(), json=data)
```
rtibbles, line 171: Not sure why this code block has been dedented? It wasn't in your other PR.

ricecooker/utils/downloader.py
```diff
@@ -207,6 +223,9 @@ def make_request(
     retry_count = 0
     max_retries = 5
     request_headers = DEFAULT_HEADERS
+
+    configure_download_session(sess, user_email)
```
rtibbles, line 227: Note, I think we should be setting this more globally on `config.DOWNLOAD_SESSION`, not just within the the downloader utility. While it is most acute when we are downloading HTML pages, it is better we do this more broadly.

Once we have authenticated against Studio with the token, we should be able to use the user email from there, rather than requiring it be explicitly passed in.

See here where we receive the username (which is the email address) once we have authenticated with Studio: https://github.com/learningequality/ricecooker/blob/develop/ricecooker/commands.py#L90

Once we have that information in hand, we can do the required DOWNLOAD_SESSION update.

rtibbles (changes requested):
Thanks for this, this is a good start!

This can be done more simply, and more globally by leveraging the information we get from authenticating the user against the Studio server. Please ask any follow up questions about the directions I have suggested.

Also, please make these commits independently of the commits from your other PR.

ricecooker/utils/downloader.py
Divyanshi750: Yes, passing an email manually every time can be tedious. I will work on an approach and make sure that the code is more of Don't Repeat Yourself and improves overall authentication and session management. But, I would like to ask if there any specific requirements or constraints for setting the download session that I need to take care of?

Divyanshi750:
Improved session configuration by adding global email setup after Studio authentication. 

Changes made:
- Added AuthenticationError class in exceptions.py for better error handling in case the authentication fails
- Updated download session configuration to use authenticated user's email when available
- Moved session configuration logic to be more globally accessible

This makes the download session configuration more consistent and eliminates the need 
to manually pass email for each download operation as requested.Please look into the changes suggested and review the changes made. @rtibbles

.gitignore
```diff
@@ -136,3 +136,4 @@ video_cache_py3.sqlite
 cache.sqlite
 
 chefdata/
+audio_cache.sqlite
```
rtibbles, line 139: This has been added in develop from your other PR - you are still committing the file itself here as well too, so that needs to be removed from the commit history.

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
rtibbles, line 1: This still needs to be removed, as does the file below.

setup.py
```diff
@@ -26,6 +26,7 @@
         "console_scripts": [
             "corrections = ricecooker.utils.corrections:correctionsmain",
             "jiro = ricecooker.cli:main",
+            "ricecooker = ricecooker.cli:main",
```
rtibbles, line 29: This doesn't need to be added.

tests/test_downloader.py
```diff
@@ -70,3 +76,66 @@ def test_archive_path_as_relative_url(self):
...
+
+    with pytest.raises(requests.exceptions.RequestException):
+        make_request(
+            unreliable_url,
+            user_email="retry_test@example.com",
+            timeout=1
+        )
+
+
+def test_performance_overhead():
```
rtibbles, line 106: What is this test doing?

tests/test_downloader.py
```diff
@@ -70,3 +76,66 @@ def test_archive_path_as_relative_url(self):
...
+
+
+def custom_email_request():
+    make_request("https://example.com", user_email="perf_test@example.com")
+
+
+baseline_time = timeit.timeit(baseline_request, number=100)
+custom_email_time = timeit.timeit(custom_email_request, number=100)
+
+assert custom_email_time - baseline_time < 0.01
```
rtibbles, line 123: Why are we making assertions in the module level code? This looks like it was generated by an LLM and unthinkingly pasted in here. Please focus only on testing the code you have added around updating the user agent.

ricecooker/utils/__init__.py
```diff
@@ -0,0 +1,10 @@
+import ricecooker
+
+
+def configure_download_session(session, user_email=None):
```
rtibbles, line 4: We should add this into a specific module, not in the __init__.py.

ricecooker/utils/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+    import ricecooker
```
rtibbles, line 5: Why is this being imported twice?

ricecooker/utils/downloader.py
```diff
@@ -28,6 +28,8 @@
 from ricecooker.utils.html import download_file
 from ricecooker.utils.html import replace_links
 from ricecooker.utils.zip import create_predictable_zip
+import logging
```
rtibbles, line 31: Why has logging been added here?

ricecooker/commands.py
```diff
@@ -21,6 +25,25 @@
     pass
 
 
+def authenticate_studio(username, password):
+    """
+    Studio authentication method that also configures the download session.
+    """
+    try:
+        auth_response = authenticate_studio(username, password)
```
rtibbles, line 33: Why is this function making a recursive call here?

ricecooker/utils/downloader.py
```diff
@@ -48,6 +50,14 @@
 }
 
 
+    
+from . import configure_download_session
+def download_file(url, clear_cookies=False, headers=None, timeout=60,
```
rtibbles, line 55: What is this new method for?

ricecooker/utils/downloader.py
```diff
@@ -197,7 +207,7 @@ def read(
 
 
 def make_request(
-    url, clear_cookies=False, headers=None, timeout=60, session=None, *args, **kwargs
+    url, clear_cookies=False, headers=None, timeout=60, session=None, user_email=None, *args, **kwargs
```
rtibbles, line 210: I don't think we need to pass this in, we instead copy the user agent from the globally configured DOWNLOAD_SESSION that should have had its user agent headers set properly. So far this PR is not doing this though, so is missing the main point of this issue.

rtibbles (changes requested):
In spite of tests being present, the code is full of issues that are evident from a code read through.

This has the appearance of being generated with a lot of LLM assistance, and not sufficient critical reflection of the generated code.

It is also not fixing the issue in the life cycle of a SushiChef class being executed. The header needs to be set on the central DOWNLOAD_SESSION in the `config` module - and should be done in the place where the user has already been authenticated with a token with the Studio server.

rtibbles:
The majority of comments here are still unaddressed - the main approach needs to be tweaked to focus on the config.DOWNLOAD_SESSION and to be triggered when the tree manager authenticates to Studio.

ricecooker/utils/downloader.py
rtibbles: Hi @Divyanshi750 - I'm sorry, I don't understand what isn't clear here - I have outlined which session and needs to be updated, and also shown where in the code this should happen.

rtibbles:
This has been force pushed - but there are still conflicts to be resolved, and the principle feedback given is still unaddressed.

Divyanshi750:
> This has been force pushed - but there are still conflicts to be resolved, and the principle feedback given is still unaddressed.

Hi @rtibbles , I was thinking to approach the problem in this way:
Implementing a SessionManager class directly in the tree.py file. It would handle HTTP sessions for downloads and API requests. And then create function to get the current download session or create a new one.
Inside TreeMAnager class, can we implement a function to authenticate to Studio using the provided token and initialize download session.
In downloader.py, a function that would download a file from a URL and save it to a destination or return the contents. And then return the downloaded file.
Let me know if any refinements are needed!

rtibbles:
We just need to update the session on the core config object, and have that update when authentication has happened with Studio, so we have the registered email address. That's all that is needed here.

rtibbles:
This PR has not been updated in over a month, closing.

PR #564, opened 2025-01-24
Upgrade flake8 hook for Python 3.12 compatibility.

rtibbles:
## Summary
Upgrades flake8 so that our linting checks work again, after ubuntu latest moved to Python 3.12 by default

Updates the skip check for Python tests, so that required checks show as passing even if the tests are being skipped.


rtibbles:
Merging to unblock other work here.

PR #566, opened 2025-01-25
Add webm support to compress_video. Add tests. Raise error if webm passed to faststart.

rtibbles:
## Summary
Adds support to our compress_video utility function for compressing webm videos (previously the use of mp4 specific flags would cause an error).
Makes the faststart flag function throw an error if used with webm, as webm does not have a faststart flag
Adds tests.

### DIFF
--- a/tests/media_utils/test_videos.py
+++ b/tests/media_utils/test_videos.py
@@ -251,6 +251,44 @@ class Test_compress_video:
             with pytest.raises(videos.VideoCompressionError):
                 videos.compress_video(bad_video.name, vout.name, overwrite=True)
 
+    def test_default_compression_works_webm(self, high_res_video_webm):
+        with TempFile(suffix=".webm") as vout:
+            videos.compress_video(high_res_video_webm.name, vout.name, overwrite=True)
+            width, height = get_resolution(vout.name)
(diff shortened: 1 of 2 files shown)


tests/media_utils/test_videos.py
```diff
@@ -251,6 +251,44 @@ def test_raises_for_bad_file(self, bad_video):
             with pytest.raises(videos.VideoCompressionError):
                 videos.compress_video(bad_video.name, vout.name, overwrite=True)
 
+    def test_default_compression_works_webm(self, high_res_video_webm):
+        with TempFile(suffix=".webm") as vout:
+            videos.compress_video(high_res_video_webm.name, vout.name, overwrite=True)
```
marcellamaki, line 256: are we always overwriting when we do this? i'm asking because for all of the tests overwrite is set to true. and I don't know if that's relevant but it just was notable to me

rtibbles: Not always - we only do this for convenience of using a tempfile for the output (and because the TempFile context manager creates the file, we have to allow overwrite).

marcellamaki (approved):
Code makes sense, and manual QA running locally in python shell does seem that compression works as expected

PR #567, opened 2025-01-28
added insufficient storage exception

Divyanshi750:
## Summary

This pull request solves the problem of getting an informative error if users run out of storage space; it provides custom InsufficientStorageException for uploading where the status code 412 Precondition Failed will be caught, and this will then be handled appropriately in the main application.

Specifically, I've made the following changes:

Defined the InsufficientStorageException class that inherits from the base Exception class

In the do_file_upload method of the ChannelManager class, added a check for the 412 status code and raised the InsufficientStorageException if encountered

In the _handle_upload method, propagated the InsufficientStorageException up the call stack so that it can be caught and handled in the main application

## References

Original issue: https://github.com/learningequality/ricecooker/issues/220
Original PR: https://github.com/learningequality/ricecooker/pull/562

## Reviewer guidance

To test these changes you can simulate the 412 precondition failed response by mocking the behavior of the file upload endpoint returned by the post() method config.SESSION to check whether the main application is raised and can correctly catch the InsufficientStorageException.
I have closed the older PR and raised a new one due to some technical difficulties. I have removed changes to other files and added audio-cache file to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,4 @@ video_cache_py3.sqlite
 cache.sqlite
 
 chefdata/
+audio_cache.sqlite
--- a/ricecooker/managers/tree.py
+++ b/ricecooker/managers/tree.py
@@ -144,6 +144,10 @@ class ChannelManager:
                 if not exists
             ]
 
+class InsufficientStorageException(Exception):
+    """Raised when there is not enough storage space."""
+    pass
+   
     def do_file_upload(self, filename):
         file_data = self.file_map[filename]
         if file_data.skip_upload:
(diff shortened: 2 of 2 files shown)


Divyanshi750:
Addition of insufficient storage exception has been updated in the PR. Please look into this and review it. @rtibbles.

rtibbles (approved):

rtibbles (approved):
Adds an explicit message that should break out of the upload loop to give a better warning.

Divyanshi750:
Thank you for your time, review and suggestions.

Issue #558, opened 2025-01-09
Enhanced spreadsheet/CSV workflow


rtibbles:
## Overview 

Prototype and create an enhanced spreadsheet/CSV based import flow that specifies enhanced metadata, encoded using human readable language and with validation.

#### Description and outcomes
* Read from CSVs, Excel Spreadsheets or Google Spreadsheets
* An example input spreadsheet can be seen here: https://docs.google.com/spreadsheets/d/1ppemnjfeOL-HrlUqJMgur3GQ4nIZOdqyMLRJa4O9vkY/edit?gid=2064123036#gid=2064123036
* Add handling for DownloadFile to download from Google Drive links
* Create a new ContentNode subclass that will turn itself into the appropriate ContentNode class depending on the kind of file using file identification sniff checks
* Initially all metadata lookups can be handled via the associated sheet - but this could eventually be a core part of the feature

Kriti382:
Hey! I would like to work on this issue.

ozer550:
> Hey! I would like to work on this issue.

Hi @Kriti382! This issue is already assinged. But there are contributing opportunities in several repositories. See the contributing guidelines including links to issues suitable for contribution for each repository here:

- [Kolibri Contributing Guidelines](<https://github.com/learningequality/kolibri/blob/release-v0.16.x/CONTRIBUTING.md>)
- [Kolibri Design System Contributing Guidelines](<https://github.com/learningequality/kolibri-design-system/blob/main/CONTRIBUTING.md>)
- [Studio Contributing Guidelines](<https://github.com/learningequality/studio/blob/unstable/CONTRIBUTING.md>)

You can also see the list of unassigned ["help wanted"](<https://github.com/search?q=org%3Alearningequality+is%3Aopen+is%3Aissue+label%3A%22help+wanted%22+no%3Aassignee&type=issues>) or ["good first issue"](<https://github.com/search?q=org%3Alearningequality+is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22+no%3Aassignee&type=issues>) issues across all repositories.

PR #570, opened 2025-02-11
Add utility to validate stream of audio and video media files.

rtibbles:
## Summary
Adds a utility to validate that audio and video files are valid and readable

## References
Used in work for #558


marcellamaki (approved):
after a slight adjustment to the code comments to clarify the errors, this looks good to go

PR #571, opened 2025-02-15
Adds more tests for the File classes to ensure behaviour ahead of refactor

rtibbles:
## Summary
* Tweaks test to cleanup file that is cloned for a test
* Deletes a bunch of tests that simply did `assert True` and were doing nothing of any use
* Adds tests for File class and subclass behaviours
* Skips some tests that show what they _should_ do, but currently fail to.
* These tests will act as a spec for some of the updates in subsequent refactor.

## References
In support of #558

## Reviewer guidance
Do all tests make sense? Do they pass on all platforms for all Python versions?


marcellamaki (approved):
since this is only adding tests, and the code is very readable, i will approve. we can adjust (or make further additions) to tests as needed in the future in the refactor if different behavior is needed

PR #572, opened 2025-02-18
Update graphie files to focus on https only downloads.

rtibbles:
## Summary
* Updates graphie tests to focus on downloading from https, rather than local file paths

### DIFF
--- a/ricecooker/classes/files.py
+++ b/ricecooker/classes/files.py
@@ -1243,7 +1243,7 @@ class _ExerciseGraphieFile(DownloadFile):
     default_ext = file_formats.GRAPHIE
 
     def __init__(self, path, **kwargs):
-        self.original_filename = path.split(os.path.sep)[-1].split(".")[0]
+        self.original_filename = path.split("/")[-1].split(".")[0]
         super(_ExerciseGraphieFile, self).__init__(path, **kwargs)
 
     def get_preset(self):
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -2,11 +2,11 @@
 import base64
 import hashlib
 import os.path
-import sys
 import tempfile
 import zipfile
 from io import BytesIO
 from shutil import copyfile
+from unittest.mock import MagicMock
 from unittest.mock import patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


tests/test_files.py
```diff
@@ -1182,43 +1182,60 @@ def test_different_images():
...
+                content = SAMPLE_JSON.encode("utf-8")
+            else:
+                raise Exception(f"Unexpected URL pattern: {mock_response.url}")
+
+            if chunk_size is None:
+                yield content
+                return
+
+            # Yield content in chunks of specified size
+            for i in range(0, len(content), chunk_size):
```
marcellamaki, line 1205: I don't understand which scenarios we would actually be passing in a (test?) chunk_size here. I think this is just in different files? i.e. `chefs.py` https://github.com/marcellamaki/ricecooker/blob/develop/ricecooker/chefs.py#L859

marcellamaki (approved):
No blockers but just some clarifying questions. 

Also wondering - is there any utility to having tests on a different test path? beyond something like the pattern:
`https://example.com/graphie/test-graph` 

I am not nearly familiar enough with graphie files to know if having test cases for any variations on this is worthwhile. (Beyond with the .svg and .json file types I mean. Different or potentially more complex? path structure)

ricecooker/classes/files.py
```diff
@@ -1243,7 +1243,7 @@ class _ExerciseGraphieFile(DownloadFile):
     default_ext = file_formats.GRAPHIE
 
     def __init__(self, path, **kwargs):
-        self.original_filename = path.split(os.path.sep)[-1].split(".")[0]
+        self.original_filename = path.split("/")[-1].split(".")[0]
```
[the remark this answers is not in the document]
rtibbles: Correct, and so on windows, this actually breaks things, because `os.path.sep` is `\` there.

tests/test_files.py
rtibbles: The content is always going to be smaller than the chunk size - I just wanted to make sure we were iterating over it, just in case we wanted to make it bigger at some point.

rtibbles:
> Also wondering - is there any utility to having tests on a different test path? beyond something like the pattern:

You'll not that I do use different paths for some tests - specifically to side step the caching that happens in other tests.

PR #573, opened 2025-02-20
Pipeline refactor

rtibbles:
## Summary
* Refactors the vast majority of file handling into a staged pipeline that can take any path as input and run appropriate processing on the file.
* Three stages at the moment are Download, Convert/Validate, and Extract Metadata. 
* Adds validation for audio and video files and zip based file types. 
* Adds a mostly unused extract metadata stage that mostly infers presets for now, but could extract additional metadata in future. 
* The pipeline and stages are extensible, so thumbnail generation could be added later as an additional stage as well as integration with other tooling to generate descriptions, subtitles etc. 

## References
Prerequisite work for #558

## Reviewer guidance
Do all the tests pass?
Do the places where the tests have been changed make sense?
Does the overall structure of the pipeline feel understandable?

N.B. I have deliberately skipped the HTML5 dependency test case, which currently fails, as the example zip does not have an index.html.

We will need some slightly more nuanced validation to allow this to pass in future - but as this is very rarely used this seems acceptable for now.

See: https://github.com/search?q=org%3Alearningequality%20HTML5_DEPENDENCY_ZIP&type=code

Only one instance is of actual active code in a content integration script - one other is from commented out code using the archived imscp project, and the others are either ricecooker, le-utils, or one of these in an inadvertently committed virtualenv.

Further work on this can happen when this PR is finalized: https://github.com/learningequality/ricecooker/pull/468

### DIFF
--- a/ricecooker/utils/pipeline/__init__.py
+++ b/ricecooker/utils/pipeline/__init__.py
@@ -1,0 +1,75 @@
+from copy import deepcopy
+from typing import Dict
+from typing import Optional
+
+from .convert import ConversionStageHandler
+from .extract_metadata import ExtractMetadataStageHandler
+from .file_handler import CompositeHandler
+from .transfer import DownloadStageHandler
+from ricecooker.utils.pipeline.context import FileMetadata
+
+
+class FilePipeline(CompositeHandler):
+    """
+    A class to manage a sequence of handlers and execute them in order.
+    Each handler should be a subclass of Handler.
+    The pipeline object will store global context that will be passed to each handler,
+    but will be overridden by the context generated during the course of a file's processing.
+
+    This pipeline can be customized by passing `children` as an argument to the constructor.
+
+    For example to add a custom stage to the pipeline, you can do:
+    ```python
+    from ricecooker.utils.pipeline import FilePipeline
+    from ricecooker.utils.pipeline.custom_stage import CustomStageHandler
+    pipeline = FilePipeline(children=[CustomStageHandler()])
+    ```
+
+    To just modify one of the existing stages, you can do:
+    ```python
+    from ricecooker.utils.pipeline import FilePipeline
+    from ricecooker.utils.pipeline.convert import ConversionStageHandler
+    from ricecooker.utils.pipeline.extract_metadata import ExtractMetadataStageHandler
+    from ricecooker.utils.pipeline.transfer import DownloadStageHandler
+    from ricecooker.utils.pipeline.transfer import DiskResourceHandler
+
+    download_stage = DownloadStageHandler(children=[DiskResourceHandler()])
+    pipeline = FilePipeline(children=[download_stage, ConversionStageHandler(), ExtractMetadataStageHandler()])
+    ```
+    This will replace the default `DownloadStageHandler` with a new one that has a `DiskResourceHandler` as its only child.
+    """
+
+    DEFAULT_CHILDREN = [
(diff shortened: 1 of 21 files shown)


ricecooker/utils/pipeline/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+class FilePipeline(FilePipelineBase):
+    """
+    The default file pipeline for ricecooker usage.
+    """
+
+    CHILDREN = [
+        DownloadStageHandler,
+        ConversionStageHandler,
+        ExtractMetadataStageHandler,
+    ]
```
bjester, line 16: I was thinking about this more and the workaround you explained for passing options to handlers like the video compression handler. I think that raises the idea that overall, the architecture could benefit from a looser design, which uses a factory to construct these, rather than statically defining them. That would allow for better composition and customization of different handlers.

For example, in Morango, something like that isn't really necessary for the operations because their role is very rigid. In curriculum automation, that flexibility is useful, and so I used a factory pattern to define the 'default' behavior and setup. The combination provides both the ready-to-run approach and the flexibility to adapt or add later on without explicitly modifying the existing code.

rtibbles: In an earlier iteration, I deferred instantiation of the handler classes, and iterated over the classes instead of the instantiated objects. Then each time a handler was instantiated it was with a specific path attribute and context object.

In that way, the pipeline was kind of its own factory.

I do think being able to add children dynamically to the StageHandler class and Pipeline class would be helpful (at init), so I can tweak that somewhat.

Have a factory function for each stage, with a default, and then an overall default pipeline that invokes those factory functions too.

I am also thinking about how I can surface the requisiteness of some parts of the context earlier in the pipeline, and I think I have a way to do it that should not be too onerous.

We already have the tree structure of each stage to play with, and for any particular path that gets input, if we can just report what kind of file will be outputted by it, then we can infer its entire path through the pipeline, and then warn early if arguments are missing.

I will implement both of these and do a force push.

rtibbles: Actually, thinking it through a bit more, the 'output prediction' is going to get squirrelly fast, especially when I add Google Drive download which could be hiding any file type behind its URL. So I'll just focus on the factory functions.

rtibbles: Updated!

marcellamaki (approved):
Co-reviewed with Richard and then re-reviewed separately. 

Overall, the approach seems like a good plan in terms of the management of responsibilities. Richard will address Blaine's remaining feedback about ffmpeg/flexibility in follow up. 

Since there is pre-existing work adding in additional test coverage, we can be fairly confident that the behavior has been retained in the refactor.

tests/pipeline/test_transfer.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+from ricecooker.utils.pipeline.transfer import (
+    get_filename_from_content_disposition_header,
+)
+
+
+content_disposition_filename_cases = [
+    ('Content-Disposition: attachment; filename="example.jpg"', "example.jpg"),
```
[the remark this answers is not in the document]
rtibbles: Probably? This test was pre-existing, I just moved it.

rtibbles:
Follow up issue to make the pipeline more easily configurable: https://github.com/learningequality/ricecooker/issues/592

PR #582, opened 2025-03-07
Revert "Allow for chefs to explicitly set a channel ID."

rtibbles:
## Summary
This reverts commit 9e394e721ab8bf9b904bd60bc8dba30c0d841901.
Reverts dangerous update that allows for node_id collisions across channels.


nucleogenesis (approved):
Looked at #285 and see that the changes have been reverted as expected. Should be good to go

PR #584, opened 2025-03-07
Handle case where the file extension has changed from the original filename

rtibbles:
## Summary
* Studio ignores the file format we send and just uses the "filename" property to determine the appropriate extension for the file
* We already handled the case when the extension did not exist, this adds handling for when the file has been converted, so the original filename extension and the extension of the file we are uploading are totally different

See Slack discussion for more details: https://learningequality.slack.com/archives/C6S90FCDT/p1740676947066549

### DIFF
--- a/ricecooker/managers/tree.py
+++ b/ricecooker/managers/tree.py
@@ -167,7 +167,7 @@ class InsufficientStorageException(Exception):
             # it does not currently use the passed in file_format as the default
             # extension.
             name, ext = os.path.splitext(data["name"])
-            if not ext:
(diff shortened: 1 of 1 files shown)


marcellamaki (commented):
The code update makes sense to me, just not sure if doing something to test this would be helpful, or if we can approve from here

marcellamaki (approved):
co-reviewed [relevant studio code](https://github.com/learningequality/studio/blob/unstable/contentcuration/contentcuration/viewsets/file.py#L245C11-L245C40)

PR #593, opened 2025-04-09
Allow ContentNode class to take a uri argument for its default file

rtibbles:
## Summary
* Updates the ContentNode class to allow just passing a URI for the resource, which will then infer the kind, files etc to make the node
* Updates pipeline metadata extraction to infer kind for an associated content node
* Cleans up ContentNode behaviour to allow deferred setting of the kind

## References
Part of #558

## Reviewer guidance
Do tests cover the appropriate code paths? Are they missing anything? Does the code make sense?

## Summary by CodeRabbit

- **New Features**
  - Added support for processing content nodes using a pipeline with URI-based file handling.
  - Content nodes now automatically infer their type and learning activities based on the resource file type.
  - Enhanced metadata extraction for HTML5 files, including kind inference and metadata merging.
  - Introduced comprehensive metadata inheritance and validation across node hierarchies.
  - Added new metadata fields and improved license serialization and validation.
  - Unified file preset handling with default presets and improved thumbnail file format associations.
  - Added explicit error handling to remove invalid nodes during validation and processing.
  - Enabled optional cache skipping in pipeline and file handling stages for more flexible processing.

- **Bug Fixes**
  - Improved handling of context initialization in file pipelines to prevent errors when context is not provided.
  - Enhanced error messages for file preset retrieval and improved pipeline execution with optional cache skipping.
  - Fixed license validation error messages to be clearer and simplified cache usage control in file handling.

- **Tests**
  - Added tests to verify automatic kind and learning activity inference for various file types.
  - Updated tests to check for new cache keys related to HTML5 metadata extraction.
  - Added tests for metadata inheritance and merging in node hierarchies.
  - Updated tests to expect exceptions during file processing rather than validation for invalid content nodes.
  - Refactored thumbnail failure tests to focus on thumbnail file processing directly.
  - Updated exercise tests to expect consistent exception types for invalid extra field values.
  - Added tests covering robust error handling in node validation, processing, and upload management.

### DIFF
--- a/ricecooker/classes/nodes.py
+++ b/ricecooker/classes/nodes.py
@@ -45,6 +45,23 @@ kind_activity_map = {
 }
 
 
+inheritable_simple_value_fields = {
+    "language",
+    "license",
+    "author",
+    "aggregator",
+    "provider",
+}
+
+
+inheritable_metadata_label_fields = [
+    "grade_levels",
+    "resource_types",
+    "categories",
+    "learner_needs",
+]
+
+
 class Node(object):
     """Node: model to represent all nodes in the tree"""
 
@@ -56,7 +73,23 @@ class Node(object):
 
     def __init__(
         self,
+        source_id,
         title,
+        author="",
+        aggregator="",
+        provider="",
+        copyright_holder="",
+        license=None,
+        license_description=None,
+        tags=None,
+        domain_ns=None,
+        grade_levels=None,
+        resource_types=None,
+        learning_activities=None,
+        accessibility_labels=None,
+        categories=None,
+        learner_needs=None,
+        role=roles.LEARNER,
         language=None,
         description=None,
         thumbnail=None,
@@ -64,7 +97,11 @@ class Node(object):
         derive_thumbnail=False,
         node_modifications={},
         extra_fields=None,
+        suggested_duration=None,
     ):
+        assert isinstance(source_id, str), "source_id must be a string"
+        self.source_id = source_id
+
         self.files = []
         self.children = []
         self.descendants = []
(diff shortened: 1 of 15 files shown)


ricecooker/utils/pipeline/extract_metadata.py
```diff
@@ -32,6 +36,12 @@ def handle_file(self, path):
         preset = self.infer_preset(path)
         if preset:
             metadata["preset"] = preset
+            kind = KIND_FROM_PRESET.get(preset)
```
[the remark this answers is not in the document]
rtibbles: In this case the preset value is the id.

ricecooker/classes/nodes.py
```diff
@@ -1677,3 +1678,59 @@ def to_dict(self):
 
 # add alias for back-compatibility
 RemoteContentNode = StudioContentNode
+
+
+class AutomaticResourceNode(ContentNode):
```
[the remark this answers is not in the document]
rtibbles: Updated.

rtibbles:
To handle @coderabbitai's feedback and more, after discussion with @marcellamaki, I will be making some updates here to focus validation on the file presets, rather than the file classes.

In order to have full compatibility with the URI based mechanism, and ensuring any preset inference has happened already, this should only happen after process_files has been called.

ricecooker/classes/nodes.py
```diff
@@ -773,15 +793,27 @@ def __init__(
...
         license_description=None,
         copyright_holder=None,
         **kwargs,
     ):
         self.set_license(
             license, copyright_holder=copyright_holder, description=license_description
         )
+        self.uri = uri
+        self._pipeline = pipeline
+        # We should do additional validation of files once they have been processed
```
marcellamaki, line 849: I don't really understand this comment in this context. I know what it says, and I think "yes", but I don't fully know why it's _here_. It's because we have `self._files_processed = False` which means we will do more later?

rtibbles: Yes, it's a really rather unelucidatory comment, to the point that I thought maybe I had copy pasted it from Claude output, but honestly I don't think I did?

rtibbles: Have updated the comment to be much more explicit and (hopefully) clear.

rtibbles:
As discussed, nodes are now flagged as valid or invalid during the validation process. When errors are raised either during validation of file processing by the ChannelManager class, it stores the string representation of the error on the node under a `_error` attribute, which is then reused for reporting.

Any invalid nodes, or ones with failed file uploads, are now excluded from upload to Studio. They are reported node by node, rather than under the umbrella of their parent, this should give more inspectability of errors that are happening.

If the content integrator prefers things to error, the `config.STRICT` configuration can be used.

marcellamaki (approved):
The changes here look aligned to the coderabbit recommendations and the approach we discussed. I think this is good to go

PR #596, opened 2025-04-14
Fix header set for the DOWNLOAD_SESSION

GautamBytes:
## Summary

This PR ensures that all download requests made by the DOWNLOAD_SESSION in the ricecooker project include a properly formatted User-Agent header. The header follows the format: Ricecooker/{version} bot ({username}), aligning with Wikimedia's User-Agent policy.

## References

Addresses  #483

##Manual Verification Steps
- Executed a test script that sends a request to https://httpbin.org using the configured DOWNLOAD_SESSION.
- Verified that the response includes the correct User-Agent header:

WhatsApp Image 2025-04-14 at 17 16 22_9c00d1b4

- Confirmed that the header is correctly set for all download requests.

## Reviewer guidance

…

### DIFF
--- a/ricecooker/commands.py
+++ b/ricecooker/commands.py
@@ -39,7 +39,7 @@ def uploadchannel(  # noqa: C901
     publish=False,
     compress=False,
     stage=False,
-    **kwargs
+    **kwargs,
 ):
     """uploadchannel: Upload channel to Kolibri Studio
     Args:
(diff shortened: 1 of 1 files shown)


GautamBytes:
@rtibbles , I am sorry to raise a pr before getting officially assigned on issue but as nobody has commented there other than me so i thought to raise a pr. Can you please review it and see if it solves the issue?

ricecooker/commands.py
```diff
@@ -298,7 +301,6 @@ def upload_files(tree, file_diff):
         file_diff ([str]): list of files to upload
     Returns: None
     """
-    # Upload new files to CC
```
rtibbles, line 301: I appreciate the typo correction elsewhere - but let's leave these deleted comments as is!

ricecooker/commands.py
```diff
@@ -83,6 +83,9 @@ def uploadchannel(  # noqa: C901
         # Authenticate user and check current Ricecooker version
         username, token = authenticate_user(token)
         config.LOGGER.info("Logged in with username {0}".format(username))
+        config.DOWNLOAD_SESSION.headers.update({
```
rtibbles, line 86: This looks precisely right.

rtibbles (changes requested):
This looks like the right fix - please just revert the comment deletion.

GautamBytes:
@rtibbles , I have the linting issue and added some comments which were deleted. You can have a look now!

rtibbles (approved):
Excellent, thank you.

PR #614, opened 2025-06-23
Add Python 3.13 support

rtibbles:
## Summary
Adds Python 3.13 support
Runs tests across OSes for Python 3.13


rtibbles:
Note: we should squash merge this to give a more meaningful commit message.

PR #618, opened 2025-07-14
Add optional Sentry integration

rtibbles:
## Summary
Adds optional sentry integration for reporting unhandled errors


marcellamaki (approved):
Cross referenced with the sentry integration docs -- this seems like the correct change based on my understanding

PR #619, opened 2025-07-14
 Add optional Slack integration for channel upload notifications

rtibbles:
## Summary
- Add a new `send_slack_notification` function to `ricecooker/utils/slack.py` to send a simple text message to a Slack webhook.
- Call the `send_slack_notification` function in `ricecooker/commands.py` after a channel has been successfully uploaded.

### DIFF
--- a/ricecooker/config.py
+++ b/ricecooker/config.py
@@ -269,6 +269,16 @@ try:
 except ImportError:
     pass
 
+# Slack webhook URL for channel upload notifications
+SLACK_WEBHOOK_URL = os.getenv("SLACK_WEBHOOK_URL", None)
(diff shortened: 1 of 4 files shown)


ricecooker/config.py
```diff
@@ -260,6 +260,13 @@ def delete_temp_dir():
     tempfile.tempdir = chef_temp_dir
 
 
+# Slack webhook URL for channel upload notifications
+SLACK_WEBHOOK_URL = os.getenv("SLACK_WEBHOOK_URL", None)
```
marcellamaki, line 264: so the os environment in question here will be Vader (at least for now), right?

PR #639, opened 2025-11-11
Fix rewriting of CSS @import statements that use strings instead of url()

jaltekruse:
## Summary
While trying to import a textbook written in PreteXt I found a missing case in the code that finds referenced resources inside of CSS files. While most cases require using the url("http://abc.com") syntax, there are a few instances where bare "string" types are acceptable in the CSS spec, the one I hit was in an `@import` statement.

## References
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@import
url - Is a `<string>` or a `<url>` type representing the location of the resource to import.
       The URL may be absolute or relative.

## Reviewer guidance
I'll add an automated test, just wanted to create this PR as a place to document my work in the open and write some additional notes

### DIFF
--- a/ricecooker/utils/downloader.py
+++ b/ricecooker/utils/downloader.py
@@ -252,6 +252,16 @@ def make_request(
 
 _CSS_URL_RE = re.compile(r"url\(['\"]?(.*?)['\"]?\)")
 
+# Handle CSS imports with a string rather than a url()
+# From MDN docs on @import:
+# https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@import
+# url - Is a <string> or a <url> type representing the location of the resource to import.
+#       The URL may be absolute or relative.
+# Note that this regex is for the specialized import case that can be given a regular
+# string with a URL in it. An @import given a value of type <url>, will be handled
+# by the above regex and associated code.
+_CSS_IMPORT_RE = re.compile(r"@import\s*['\"](.*?)['\"]")
+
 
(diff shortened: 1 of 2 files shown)


jaltekruse:
While I'm working on a test, I'm also going to try to look up any other lingering cases where bare strings are accepted inside of CSS for URLs, that don't need a url() to wrap them. So far a quick search only gave me relevant info in the Gemini AI summary from google, I didn't find an independent page/discussion about this topic:

URLs in CSS can appear as plain strings in the following contexts:
@import rule: When importing another CSS stylesheet, the URL can be specified as a plain string.
Code

    @import "path/to/another.css";
This is equivalent to using the url() function:
Code

    @import url("path/to/another.css");
content property (for generated content): When using the content property with pseudo-elements (::before, ::after), a URL can be included as part of the string content.
Code

    .element::before {
      content: "Visit our website: https://example.com";
    }
cursor property (for custom cursors): A URL to an image file can be provided as a string value for a custom cursor.
Code

    body {
      cursor: url("path/to/custom-cursor.png"), auto;
    }
In other contexts where URLs are used in CSS, such as for background-image, list-style-image, border-image-source, or @font-face src, the url() functional notation is typically required to explicitly denote a URL value.

MisRob:
Hi @jaltekruse, we will have a look, thanks!

rtibbles (approved):
I made some test tweaks to simplify tests and get them to pass on Windows. One small correctness fix to allow spaces after `@import` statements.

PR #653, opened 2026-01-20
Courses, Units, Lessons, and KPUB

rtibbles:
## Summary
Adds support for uploading KPUB files
Adds Courses, Units, and Lessons nodes with validation
Adds example courses script for two sample courses.

## References
Fixes #638


bjester (commented):
Needs to update the `le_utils` version

tests/pipeline/test_convert.py
```diff
@@ -120,3 +124,165 @@ def test_archive_no_compression_when_disabled(video_file, audio_file):
...
+    temp_file.close()
+
+    try:
+        handler = KPUBConversionHandler()
+        with pytest.raises(InvalidFileException) as exc_info:
+            handler.validate_archive(temp_file.name)
+
+        assert "zip" in str(exc_info.value).lower()
+    finally:
+        os.unlink(temp_file.name)
```
bjester: I presume these are LLM generated?

6 out of 7 tests follow pretty much the same pattern, with this test being the outlier. A test case would eliminate the redundancy.

Although, the approach is somewhat poor. It could use a context manager without `delete=False` and eliminate the try / finally / unlink. With that change, a test case isn't quite the dramatic cleanup.

ricecooker/classes/curriculum.py
```diff
@@ -0,0 +1,65 @@
...
+    def validate(self):
+        """
+        Validate the learning objective against the le_utils schema.
+
+        Returns:
+            bool: True if valid, False otherwise.
+        """
+        # Text must be non-empty and not just whitespace
+        # Pattern from le_utils schema: ^\s*\S[\s\S]*$
+        if not self.text or not re.match(r"^\s*\S[\s\S]*$", self.text):
```
bjester: I believe this regex would pass with just `\t` or `\n`

ricecooker/classes/nodes.py
```diff
@@ -1567,3 +1580,257 @@ def to_dict(self):
...
+    @property
+    def questions(self):
+        """Return an iterable of just the question objects for file processing."""
+        return [q for q, _, _ in self.test_questions]
+
+    @questions.setter
+    def questions(self, value):
+        """Setter to allow base Node class initialization (ignored for UnitNode)."""
+        # UnitNode manages questions through test_questions, so ignore base class assignment
+        pass
```
bjester: I presume this is for consistency with `ExerciseNode`, but they have no shared interface, besides declaring this property.

tests/test_curriculum.py
```diff
@@ -0,0 +1,625 @@
...
+        """VARIANT_A constant exists."""
+        assert VARIANT_A is not None
+
+    def test_variant_b_exists(self):
+        """VARIANT_B constant exists."""
+        assert VARIANT_B is not None
+
+    def test_variants_are_distinct(self):
+        """VARIANT_A and VARIANT_B have different values."""
+        assert VARIANT_A != VARIANT_B
```
bjester: Surely a code review is sufficient for this 😅

tests/pipeline/test_extract_metadata.py
```diff
@@ -0,0 +1,61 @@
...
+            )
+
+            pipeline = FilePipeline()
+            results = pipeline.execute(temp_archive.name)
+            result = results[0]
+
+            assert result.content_node_metadata is not None
+            assert result.content_node_metadata["kind"] == content_kinds.DOCUMENT
+        finally:
+            os.unlink(temp_archive.name)
```
bjester: Similar to the other kpub tests

tests/pipeline/test_convert.py
rtibbles: Yes - sorry, clearly I forgot to tag this, and I am now remembering that I never came around and did another pass over the code, just pushed the draft once I had the test channels in place.

Will rejig the tests, then double check the implementation.

rtibbles:
Have reworked and deduplicated the tests here.

examples/curriculum_courses/sushichef.py
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    def construct_channel(self, **kwargs):
+        """Build channel structure from content.json."""
+        channel = self.get_channel(**kwargs)
+        content = load_content()
+
+        # Create a temp directory for KPUB files (cleaned up automatically)
+        self._temp_dir = tempfile.TemporaryDirectory(prefix="curriculum_kpub_")
+        self.temp_dir = self._temp_dir.name
```
bjester: Not a big deal since this is an example, but I'm assuming `self._temp_dir` doesn't need to be set on the class? Looks a little weird.

ricecooker/classes/curriculum.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(self, text, metadata=None):
+        """Create a new LearningObjective with deterministic UUID5 from text."""
+        self.text = text
+        self.id = uuid.uuid5(LEARNING_OBJECTIVE_NAMESPACE, text).hex
+        self.metadata = metadata or {}
+
+    def validate(self):
+        """Validate the learning objective against the le_utils schema."""
+        return bool(self.text and self.text.strip())
```
bjester: The UUID generation in the constructor should fail if `text` is `None`, so is there use for `self.text` too?

ricecooker/classes/nodes.py
```diff
@@ -1482,3 +1497,266 @@ def to_dict(self):
...
+            question: Question instance (same types as ExerciseNode)
+            variant: VARIANT_A or VARIANT_B
+            learning_objectives: List of LearningObjective instances
+        """
+        if variant not in (VARIANT_A, VARIANT_B):
+            raise InvalidNodeException("variant must be VARIANT_A or VARIANT_B")
+        if not learning_objectives:
+            raise InvalidNodeException(
+                "Question must have at least one learning objective"
+            )
```
bjester: Seems `add_child` and `add_question` both validate this and both call `_validate_learning_objectives`. Any reason not to add it there?

ricecooker/classes/nodes.py
```diff
@@ -1482,3 +1497,266 @@ def to_dict(self):
...
+        Add a LessonNode with its associated learning objectives.
+
+        Note: this intentionally has a different signature from
+        _CurriculumNode.add_child(node) — learning_objectives is required.
+        This is safe because add_child is only called during channel
+        construction by chef code, never by internal tree-walking code.
+
+        Args:
+            node: LessonNode to add
+            learning_objectives: List of LearningObjective instances
```
bjester, line 1600: Seems ricecooker has inconsistent docstring styles...

tests/test_curriculum.py
```diff
@@ -0,0 +1,656 @@
...
+        lesson = LessonNode(source_id="lesson-1", title="Lesson 1")
+        with pytest.raises(TypeError, match="learning_objectives"):
+            unit.add_child(lesson)
+
+    def test_accepts_lesson_node_with_learning_objectives(self):
+        """UnitNode accepts LessonNode with learning_objectives."""
+        unit = UnitNode(source_id="unit-1", title="Math Unit")
+        lesson = LessonNode(source_id="lesson-1", title="Lesson 1")
+        lo = LearningObjective("Understand addition")
+        unit.add_child(lesson, learning_objectives=[lo])
```
bjester: The tests here are passing this as a kwarg, which I know is okay, but I think this is considered bad practice?

tests/test_curriculum.py
```diff
@@ -0,0 +1,656 @@
...
+        # lo2: 1 in A, 1 in B = 2 total (per-variant balanced)
+        # But lo1 has 4 questions and lo2 has 2 — unequal across LOs
+        unit.add_question(make_question("q1"), VARIANT_A, learning_objectives=[lo1])
+        unit.add_question(make_question("q2"), VARIANT_A, learning_objectives=[lo1])
+        unit.add_question(make_question("q3"), VARIANT_A, learning_objectives=[lo2])
+        unit.add_question(make_question("q4"), VARIANT_B, learning_objectives=[lo1])
+        unit.add_question(make_question("q5"), VARIANT_B, learning_objectives=[lo1])
+        unit.add_question(make_question("q6"), VARIANT_B, learning_objectives=[lo2])
+
+        with pytest.raises(InvalidNodeException):
```
bjester: These tests look like good coverage, but I think it would provide more certainty that the validation cases are hitting the right spots with some exception message matching for each case.

bjester (changes requested):
Overall everything looks good enough for merge, but I think it'd be worthwhile to match exception messages in `TestUnitNodeValidation` considering the complexity of the validation. And so for that, at least, I'm requesting changes.

examples/curriculum_courses/sushichef.py
rtibbles: Yeah, not needed at all.

ricecooker/classes/curriculum.py
rtibbles: It won't fail if text is an empty or all whitespace string - but I've moved this into __init__.py to raise earlier.

ricecooker/classes/nodes.py
rtibbles: Nope - made the error message more generic and moved it.

ricecooker/classes/nodes.py
rtibbles: Yes, they're all over the shop - I think we should just decide what our standard is across our repos and then standardize.

tests/test_curriculum.py
rtibbles: Yes, updated to just pass positionally.

tests/test_curriculum.py
rtibbles: Added message matching to all to check this.

rtibbles:
Updated!

bjester (approved):
Looks good to me. All previous code review items resolved.

PR #656, opened 2026-02-17
Add AGENTS.md and CLAUDE.md for AI agent guidance

rtibbles:
## Summary

Add two documentation files to guide AI coding agents working in this repository:

- **`AGENTS.md`** — Concise, model-agnostic guidance (36 lines) with commands-first layout, critical gotchas, project overview, and key architecture. Designed to fit in smaller context windows.
- **`CLAUDE.md`** — Imports `AGENTS.md` via `@AGENTS.md` reference and adds Claude Code-specific extensions: detailed pipeline architecture, `le_utils` dependency context, and expanded file type implementation guidance.

The `AGENTS.md` content was validated through 3 rounds of TDD testing (RED-GREEN-REFACTOR) using Haiku and Sonnet subagents with a KPUB file type implementation task. Each round identified common agent mistakes and refined the gotchas to prevent them. Key gotchas that proved effective:

- Preventing agents from creating new File/Node subclasses in legacy `classes/` modules
- Requiring `le_utils` constant verification before use (names are non-obvious)
- Enforcing correct test placement in `tests/pipeline/`
- Preventing over-copying of validation logic from existing handlers

## References

Follows the AGENTS.md / CLAUDE.md split pattern from [Kolibri PR #14164](https://github.com/learningequality/kolibri/pull/14164).

## Reviewer guidance

- These are documentation-only changes (no code modifications)
- **AGENTS.md Critical Gotchas section** — Does this accurately capture the most common agent mistakes when working in this codebase? Are there other gotchas that should be added?
- **CLAUDE.md extended architecture** — Does this provide enough context for Claude Code without being redundant with AGENTS.md?
- The TDD validation found that HTML body parsing over-copying and File subclass creation (for Haiku) remain hard to prevent through documentation alone — these may need structural guardrails in the future

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,36 @@
+# AGENTS.md
+
+Guidance for AI coding agents working in this repository.
+
+## Quick Start
+
+```bash
+pip install -e '.[test,dev]'        # install (use uv pip if venv was created with uv)
+pytest                              # run all tests
+pytest tests/test_files.py          # run a single test file
+pytest -k 'test_something'         # filter by test name
+pre-commit run --all-files          # lint (the ONLY way to run linting)
+```
+
+**System dependencies:** `ffmpeg` and `poppler-utils`.
+
+## Critical Gotchas
+
+- **Linting:** Always use `pre-commit run --all-files`. Never run black, flake8, or other tools directly.
+- **Line length:** 160 characters, enforced by pre-commit.
+- **New file types require exactly two changes:** (1) a conversion handler in `convert.py`, (2) a metadata extractor in `extract_metadata.py`. That's it. Do NOT touch `classes/files.py` or `classes/nodes.py` — the existing File/Node subclasses there are legacy backwards-compatibility APIs that are NOT needed for new file types. The pipeline infers kinds and presets automatically.
+- **Test placement:** Pipeline tests go in `tests/pipeline/` — add to existing files like `test_convert.py` and `test_extract_metadata.py`. Do not create new test files.
+- **Validation logic:** Each handler implements only the validation its spec requires. Do not copy validation from other handlers (e.g., do not add HTML body parsing or empty-body checks unless the spec explicitly requires them).
+- **le_utils constants:** Always verify constants exist by running `python3 -c "from le_utils.constants import file_formats, format_presets; print(file_formats.CONSTANT_NAME); print(format_presets.CONSTANT_NAME)"`. Names are often non-obvious (e.g., `file_formats.HTML5_ARTICLE` for kpub, `format_presets.KPUB_ZIP` not `format_presets.KPUB`).
+
+## Project Overview
+
+Ricecooker converts educational content into Kolibri content channels and uploads to Kolibri Studio. Chef scripts subclass `SushiChef`, build a node/file tree, and call `main()`.
+
+## Key Architecture
+
+- **`utils/pipeline/`**: File processing pipeline — transfer → convert → extract_metadata. Reference architecture for new code.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,36 @@
...
+pytest tests/test_files.py          # run a single test file
+pytest -k 'test_something'         # filter by test name
+pre-commit run --all-files          # lint (the ONLY way to run linting)
+```
+
+**System dependencies:** `ffmpeg` and `poppler-utils`.
+
+## Critical Gotchas
+
+- **Linting:** Always use `pre-commit run --all-files`. Never run black, flake8, or other tools directly.
```
nucleogenesis, line 19: (maybe-blocking): Never ever ever? Sometimes it might be nice to have an agent have the ability to run specific test files' tests directly without running the full suite (although ricecooker's tests probably run a lot more quickly than Kolibri's). 

Would save context to avoid running the whole pre-commit suite on all files to test a single test case too.

CLAUDE.md
```diff
@@ -0,0 +1,41 @@
...
+Three stages with ordered handlers: **transfer** (download) → **convert** (compress/transform) → **extract_metadata**. Handlers implement `should_handle()` / `execute()`. Transfer uses `FirstHandlerOnly` mode; convert and metadata use `AllHandlers` mode.
+
+### le_utils Dependency
+
+`le_utils` provides constants defining the Kolibri content model: `content_kinds`, `file_formats`, `format_presets`, `licenses`, `languages`, and label taxonomies. It increasingly includes validation schemas that ricecooker should conform to.
+
+## Development Notes
+
+- The pipeline code is the reference architecture — follow its OOP patterns when writing new code.
+- Older code (especially `ricecooker/classes/`) has less test coverage. Take extra care when modifying it.
```
nucleogenesis, line 26: suggestion: What if we also include a requirement to add regression tests when making changes to "older code" (or any existing code when relevant)?

CLAUDE.md
```diff
@@ -0,0 +1,41 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+@AGENTS.md
```
nucleogenesis, line 5: suggestion/question: Does this effectively inject the AGENTS.md in this spot in the context?

AGENTS.md reads to me (as a human anyway) as supplementary to the rest of this CLAUDE.md file and could be a little confusing to read first (again, as a human) - not sure how this may or may not impact LLMs though.

nucleogenesis (commented):
Things read very well overall. I've left a few questions to discuss before approval.

AGENTS.md
nucleogenesis: I realize this is in the "linting" section - but I worry this might be taken as "law" to the agent to "Never run black, flake8, **or other tools directly**." 

Could maybe mitigate by explicitly explaining the agent should run single tests when they don't need to run the full suite or something?

rtibbles: The main issue I have seen is with agents running one specific linting check and ignoring others, hence this directive. If we see an issue with their wasting time running the full suite, happy to update, but I feel like it's only really worth doing this when we see failure modes, not in anticipation.

CLAUDE.md
rtibbles: Would be better covered in specifics in relevant issues, I think.

CLAUDE.md
rtibbles: The divide is meant to be the AGENTS.md has the core instructions, and CLAUDE.md has the additional detail that can work in Opus 4.6's 1M token context window.

nucleogenesis (approved):
Apologies for my letting my review grow stale here - my questions are satisfactorily answered! Thanks Richard

PR #673, opened 2026-06-04
Pass compression settings through the pipeline default context

rtibbles:
## Summary

Compression was driven by the `config.COMPRESS` global, which handlers consulted at processing time. Now `FilePipeline` accepts a `default_context` merged into every `execute()` call, and `SushiChef.run` builds explicit video/audio settings from the `--compress` flag and passes them through it. All handlers — standalone media and archive — compress only when explicit settings are provided.

`config.COMPRESS` is removed since nothing reads it anymore. **Breaking** for any external chef script that read `config.COMPRESS` directly; the `--compress` CLI flag itself behaves as before.

## References

Extracted from in-progress spreadsheet chef work; no linked issue.

## Reviewer guidance

`uv run --group test pytest tests/pipeline/ tests/test_files.py`

Areas worth a careful look:
- Behavior change: with `--compress`, standalone media previously compressed with ffmpeg defaults; now `run()` supplies `crf: 32` / `max_height` / `bit_rate: 96` explicitly — check those defaults are the ones we want
- `ArchiveProcessingBaseHandler.get_cache_key` now keys on settings presence rather than `config.COMPRESS` — affects which cached results get reused across runs with different flags

## AI usage

I used Claude Code to extract this change from a larger working branch and add the tests; tests were verified to fail against the old `config.COMPRESS` gating, and I reviewed the final diff.

### DIFF
--- a/ricecooker/chefs.py
+++ b/ricecooker/chefs.py
@@ -494,7 +494,16 @@ class SushiChef(object):
         self.CHEF_RUN_DATA["current_run"] = run_id
         self.CHEF_RUN_DATA["runs"].append({"id": run_id})
 
-        self.file_pipeline = FilePipeline()
+        default_context = {}
+        if self.get_setting("compress", False):
+            default_context["video_settings"] = {
+                "crf": 32,
+                "max_height": self.get_setting("video-height") or 720,
+            }
+            default_context["audio_settings"] = {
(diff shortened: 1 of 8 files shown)


rtibbles:
Superseded by #690

PR #677, opened 2026-06-11
Enable ruff isort import rules

rtibbles:
## Summary

Import-ordering enforcement was silently lost when ricecooker migrated from `reorder_python_imports` to ruff — the `I` rule category was never added to ruff's `select`, so nothing reordered or split imports. This adds `I` and configures isort to reproduce the old one-per-line style (`force-single-line`, `order-by-type = false`, i.e. case-insensitive rather than CapWords-first). ricecooker had no isort config at all, so one was added.

Three commits: the config change, a mechanical one-time reformat (37 files, `ruff check --fix`), and a new `.git-blame-ignore-revs` (didn't previously exist) with the reformat.

## References

None — no tracking issue. Part of a cross-repo pass restoring import-ordering enforcement after the uv migrations.

## Reviewer guidance

The reformat commit is purely mechanical — review the config commit for intent. The reformat is excluded from `git blame` via `.git-blame-ignore-revs`. Verify with `ruff check --select I .` (clean on the branch).

Each reformat diff was scanned for semantic import-order changes; none were found in this repo.

## AI usage

Used Claude Code to investigate why import style was no longer enforced after the uv migration, derive the ruff isort config that reproduces the old `reorder_python_imports` behaviour, and apply it. The reformat diff was reviewed for load-bearing import-order changes before committing.


bjester (approved):
LGTM

Issue #303, opened 2021-02-01
Refactor downloader.py to make more functions unit testable


kollivier:
### Description

Currently, there are a lot of programming practices, like inner functions, in `downloader.py` that prevent us from unit testing features such as URL detection and re-writing. This leads to whack-a-mole bug fixing since we don't have the resources to test all existing sources with any changes we make.

Tharun-10Dragneel:
hi is the issue is still open?i would like to work on this issue!

jaltekruse:
@Tharun-10Dragneel I am planning to work on #655 Which is a generalization of this type of problem in the ricecooker code, and is now marked as superseding this issue. If you would like to contribute to that effort or follow along we could maybe connect on the project slack instance. I didn't see anyone named Tharun on there, if you haven't asked to join you can check out this page for getting started info on contributing https://learningequality.org/contributing-to-our-open-code-base/

rtibbles:
Has been resolved by #690

PR #690, opened 2026-07-11
Download external resources for archive content types (#233)

### LINKED ISSUE #233: Handle external URL references in archives to enable offline use in Kolibri
### Description

Archive-based content types (H5P, HTML5 zip, IMSCP) may contain references to external URLs — images, videos, fonts, stylesheets, scripts — that won't be available in offline Kolibri deployments. The pipeline conversion handlers need to scan archive contents for these references, download the resources, bundle them into the archive, and rewrite the references to point to the local copies.

#### Existing logic in `downloader.py`

We already have most of this URL extraction and rewriting logic in `downloader.py`. `download_static_assets()` and its inner functions extract URLs from:
- HTML attributes: `img[src]`, `link[href]`, `script[src]`, `source[src]`, `img[srcset]`, `[style*="background-image"]`
- CSS: `url()` references for fonts, images, etc.
- Recursive resource following (CSS that references fonts, etc.)

And `ArchiveDownloader` downloads pages with all their resources, rewrites paths, and creates ZIP archives.

However, this logic is **not reusable by the pipeline** because:
- It's tightly coupled to web scraping (takes URLs, not archive contents)
- The extraction/rewriting logic is buried in inner functions that can't be unit tested (see #303)
- It has known bugs being fixed in #636 and #639

#### kolibri-zip as reference spec

Kolibri's [`kolibri-zip`](https://github.com/learningequality/kolibri/blob/release-v0.19.x/packages/kolibri-zip/src/index.js) package handles the *runtime* side of this: when rendering ZIP-based content, it extracts files, resolves internal path references, and rewrites them to blob URLs. Its [`fileUtils.js`](https://github.com/learningequality/kolibri/blob/release-v0.19.x/packages/kolibri-zip/src/fileUtils.js) provides a comprehensive spec for which reference types need handling:

**HTML/XML files** (`src`, `href`, `srcset`, inline `style`, `<style>` blocks):
```html
<img src="images/photo.jpg">
<link href="styles/main.css">
<script src="https://cdn.example.com/lib.js"></script>
<img srcset="img-300.jpg 300w, img-600.jpg 600w">
<div style="background: url('bg.png')">
```

**CSS files** (`url()`, `@import` — both `url()` and bare string forms):
```css
@import 'fonts/custom.css';
background-image: url('../images/bg.png');
@font-face { src: url('https://fonts.example.com/font.woff2'); }
```

**H5P JSON** (`path` attributes in `content/content.json`):
```json
{
  "video": {
    "files": [
      { "path": "https://h5p.org/sites/default/files/h5p/iv.mp4", "mime": "video/mp4" }
    ]
  }
}
```

kolibri-zip handles *internal* references at runtime, but cannot fetch *external* URLs (especially offline). That's ricecooker's job at import time.

### Key architectural constraint

External resource downloading and reference rewriting must happen **before** `create_predictable_zip` is called in `ArchiveProcessingBaseHandler.handle_file()` ([convert.py](https://github.com/learningequality/ricecooker/blob/develop/ricecooker/utils/pipeline/convert.py)). Once `create_predictable_zip` runs, the archive is sealed — it iterates existing files and can only transform them via `file_converter` (currently used for media compression), not add new ones.

The flow in `handle_file()` would become:

```
validate_archive(path)
path = download_and_rewrite_external_refs(path)   # <-- new step
create_predictable_zip(path, file_converter=...)   # existing step
```

Where `download_and_rewrite_external_refs` would:
1. Extract the archive to a temp directory
2. Scan text-based files for external URL references
3. Download those resources into the temp directory
4. Rewrite references in the source files to point to local copies
5. Return the temp directory path (which `create_predictable_zip` already accepts — it handles both directories and zip files)

### Related issues and PRs

- **#303** — Refactor `downloader.py` to make URL detection and rewriting unit-testable. **This issue supersedes #303** — the extraction of URL logic into shared utilities serves both the testability goal and the pipeline integration goal.
- **#636** — Fix CSS `link` tag filtering (`"rel" in node` vs `"rel" in node.attrs`) and extensionless URL path placement. By @jaltekruse.
- **#639** — Fix CSS `@import` with bare strings (not wrapped in `url()`). By @jaltekruse.

#### Phase 1: Land bug fixes from #636 and #639

Merge the open PRs first to preserve @jaltekruse's contributor attribution in git history before the refactor changes the code structure. These fix real bugs in the URL extraction logic that the shared utilities will need to carry forward.

#### Phase 2: Extract shared utilities from `downloader.py` (supersedes #303)

Extract the URL extraction and rewriting logic from `download_static_assets()` inner functions into standalone, testable utility functions that operate on file contents (strings/bytes) rather than requiring a live HTTP session:

- **URL extraction**: Given HTML/CSS/JSON content, return a list of referenced URLs
- **URL rewriting**: Given content and a URL mapping (old → new), return rewritten content
- **External URL filtering**: Distinguish external (http/https) from internal (relative paths already in archive) references

This directly addresses #303's concern about untestable inner functions. The extracted functions can be unit tested with plain strings — no HTTP server, no filesystem, no platform-specific path issues. This also resolves the Windows test failures in #636, since the core logic tests won't depend on filesystem paths.

#### Phase 3: Create archive processing utility

Build on the Phase 2 utilities to create an archive-level processor:
- Open an archive and iterate its text-based files (HTML, CSS, JSON, XML)
- Use Phase 2 extractors to find external URL references
- Download external resources into the extracted archive directory
- Use Phase 2 rewriters to update references to local paths
- Loop detection for recursive references (à la kolibri-zip's `visitedPaths`)

#### Phase 4: Integrate into pipeline conversion handlers

Wire the archive processor into the existing handlers, running before `create_predictable_zip`:
- **`H5PConversionHandler`**: Scan `content/content.json` for external `path` values, plus HTML/CSS in content (highest priority — videos and images commonly external)
- **`HTML5ConversionHandler`**: Scan HTML/CSS files for external references
- **`IMSCPConversionHandler`**: Scan entry point HTML files and their CSS for external references

### Reference types to handle

From kolibri-zip's `fileUtils.js` and existing `downloader.py` logic:

| File type | Reference patterns | Source |
|---|---|---|
| HTML/XML | `src`, `href`, `srcset` attributes; inline `style`; `<style>` blocks | kolibri-zip `DOMMapper`, downloader.py `download_static_assets()` |
| CSS | `url()`, `@import` (both `url()` and bare string forms) | kolibri-zip `CSSMapper`, downloader.py `_CSS_URL_RE`, PR #639 `_CSS_IMPORT_RE` |
| H5P JSON | `path` attributes in `content/content.json` | H5P-specific |

### References

- kolibri-zip source: https://github.com/learningequality/kolibri/tree/release-v0.19.x/packages/kolibri-zip/src
- Current pipeline handlers: [`ricecooker/utils/pipeline/convert.py`](https://github.com/learningequality/ricecooker/blob/develop/ricecooker/utils/pipeline/convert.py)
- Existing download logic: [`ricecooker/utils/downloader.py`](https://github.com/learningequality/ricecooker/blob/develop/ricecooker/utils/downloader.py)

rtibblesbot:
## Summary

Archive-based content types (HTML5 zip, H5P, EPUB/KPUB/Bloom) can reference external URLs — images, videos, fonts, stylesheets — that do not resolve in offline Kolibri deployments. The archive conversion handlers now scan archive contents for external references, download the resources into the archive, and rewrite the references to the local copies before the archive is sealed by `create_predictable_zip`.

The URL detection and rewriting logic lives in a new module (`references.py`) with no HTTP or filesystem coupling, so it is unit-testable directly. It is organized as per-file-type `ReferenceMapper` classes — an HTML and a CSS mapper ship as the default set — modeled on kolibri-zip's per-file-type mappers. HTML rewriting is surgical: a stdlib `HTMLParser` walk collects reference spans and `_apply_edits` splices in the replacements, leaving the rest of the document byte-for-byte.

A `utils/` utility (`archive_assets.py`) drives the archive-level work. `ArchiveProcessor` takes an extracted directory and makes two passes: a reference-led walk that routes every external reference through the running file pipeline (download → convert), places the result next to its referencing file and rewrites the reference; then a scoped pass that recompresses every media file it can. The base archive handler declares the HTML and CSS mappers (`DEFAULT_MAPPERS`), so every archive format is scanned; H5P extends that set with an H5P-specific mapper for its `content.json` references.

Each reference is processed through the **running** pipeline — the handler reaches it via a new `Handler.get_pipeline()` that walks up the parent chain — so an archive asset gets the same download → convert path, config and caches as every other file, with no module-global stage. Routing archive downloads through the pipeline made the old `downloader.py` redundant: it is deleted (~1000 lines), and its two remaining live callers — the module-level `read()` and the YouTube thumbnail fetch — are migrated onto the pipeline (`read()` now wraps `DownloadStageHandler`, and the thumbnail is transcoded by the CONVERT stage from its actual bytes instead of Content-Type sniffing). The `loadjs` selenium/pyppeteer render path is intentionally dropped as a follow-up.

Compression is no longer driven by a global flag: `config.COMPRESS` is removed, and `SushiChef.run` derives explicit ffmpeg settings from `--compress` and threads them through the pipeline's new `default_context` (folding in #673), so standalone and in-archive media compress consistently.

## References

Closes #233. Supersedes #303 and #673. Carries forward the CSS fixes from #636 and #639. Follow-up filed as #692 (rename the `utils.utils` module).

## Reviewer guidance

- `ricecooker/utils/references.py:212` (`_map_html_urls`) — confirm the walk catches every reference form real HTML5 apps use (`src`, `link[href]`, `srcset`, inline `style`, `<style>` blocks) so none are silently missed.
- `ricecooker/utils/archive_assets.py:145` (`_fetch_into`) — assets are co-located with their referrer under a content-hashed basename; confirm that cannot collide across files sharing a directory.
- `ricecooker/utils/pipeline/convert.py:215` (`handle_file`) — the archive unzips into a `tempfile.TemporaryDirectory()`; confirm nothing downstream retains a path into that dir past the `with` block, so untrusted content cannot leak.

## AI usage

Used Claude Code to implement a pre-approved plan with test-driven development, writing the pure reference utilities, the archive processor, and the handler integration, then addressing review feedback across several rounds. Verified with the pipeline, references and files test suites and prek lint.

---

</details>

---

</details>

<table><tr><td><a href="https://tibbs.io/status">@rtibblesbot</a></td><td><h3>🟡 Waiting for feedback</h3><em>Last updated: 2026-07-17 18:59 UTC</em></td></tr></table>


ricecooker/utils/pipeline/convert.py
```diff
@@ -167,6 +168,13 @@ class ArchiveProcessingContextMetadata(ContextMetadata):
 class ArchiveProcessingBaseHandler(ExtensionMatchingHandler):
     CONTEXT_CLASS = ArchiveProcessingContextMetadata
 
+    # Per-handler opt-in for downloading external resource references and
```
rtibbles: So... if it's meant to be opt in - why have we defined it on the base class? Why does the base class know _anything_ about "H5P_JSON"? That should be delegated to the inheriting classes not encoded on the base class.

Did you look at the kolibri-zip implementation at all? It has a clear architecture for allowing opt in and separation of concerns here. You've just bunged everything in the base class and called it a day.

rtibbles (changes requested):
This needs to be completely rearchitected, following the exemplar referenced in the issue of the kolibri-zip module in the learningequality/kolibri repository.

Only the H5P handler needs to know _anything_ about H5P so it should be the only thing concerned with this. Use file type specific mappers and an extensible architecture to make this properly.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Good call — the base class no longer knows anything about H5P. I followed kolibri-zip's Mapper design:

- Added a `ReferenceMapper` abstraction — each file type gets a mapper that knows which files it `handles()`, what references they `extract()`, and how to `rewrite()` them.
- `references.py` now exposes only the generic web mappers (`HTMLMapper`, `CSSMapper`) as `DEFAULT_MAPPERS`.
- `H5PContentMapper` — and all knowledge of H5P, including the `content/content.json` path and `path`-key walking — lives with the `H5PConversionHandler` in `convert.py`. It opts in via `REFERENCE_MAPPERS = DEFAULT_MAPPERS + (H5PContentMapper(),)`.
- The base handler now carries only `REFERENCE_MAPPERS = ()` and the archive processor dispatches through `mapper.handles/extract/rewrite`, so it is fully format-agnostic. The `SCAN_HTML/SCAN_CSS/SCAN_H5P_JSON` flags are gone.

New file types are now supported by supplying a mapper, not by editing the base class or the processor. Done in 2b8d02f.

tests/pipeline/test_convert.py
```diff
@@ -247,3 +251,347 @@ def test_invalid_zip(self):
...
+
+                # The woff2 reference inside the downloaded CSS was rewritten.
+                css_name = next(n for n in names if n.endswith(".css"))
+                downloaded_css = open(os.path.join(static, css_name)).read()
+                assert "https://ex.com/f.woff2" not in downloaded_css
+            finally:
+                shutil.rmtree(out_dir, ignore_errors=True)
+
+    def test_h5p_json(self):
+        from ricecooker.utils.pipeline import archive_assets
```
rtibbles: Please clean up all unnecessary inline imports.

rtibbles (changes requested):
inline imports

tests/pipeline/test_convert.py
rtibblesbot: Done in 69a10b0. Hoisted every inline import in this file to the module-level import block — `archive_assets` (was duplicated across four tests), `config`, `EPubFile`, and the `BloomConversionHandler`/`EPUBConversionHandler`/`KPUBConversionHandler` imports. None had a circular-import justification. `uvx prek run --all-files` and the full `tests/pipeline/test_convert.py` suite (28 passed, 2 skipped) are green.

ricecooker/utils/downloader.py
```diff
@@ -250,19 +252,6 @@ def make_request(
     return None
 
 
-_CSS_URL_RE = re.compile(r"url\(['\"]?(.*?)['\"]?\)")
```
rtibbles, line 253: I am rather surprised most of this file hasn't been eliminated and just turned into a thin wrapper around the new logic. Any reason why not?

ricecooker/utils/pipeline/archive_assets.py
```diff
@@ -0,0 +1,201 @@
...
+import hashlib
+import mimetypes
+import os
+import tempfile
+import zipfile
+from collections import deque
+from urllib.parse import urlparse
+
+from ricecooker.config import LOGGER
+from ricecooker.utils.downloader import make_request
```
rtibbles: Why are we using some crappy "make_request" function here, when we literally have an entire pipeline that is dedicated to getting files from their source reference and properly converting them.

ricecooker/utils/archive_assets.py
```diff
@@ -0,0 +1,201 @@
...
+                abspath = os.path.join(root, name)
+                mapper = self._mapper_for(abspath)
+                if mapper is not None:
+                    worklist.append((abspath, mapper))
+        return worklist
+
+    def _process_file(self, source_path, mapper):
+        """Download ``source_path``'s external refs and rewrite them in place."""
+        try:
+            with open(source_path, encoding="utf-8") as fh:
```
rtibbles, line 118: Copy file? File pipeline does this.

ricecooker/utils/pipeline/convert.py
```diff
@@ -198,24 +208,41 @@ def handle_file(self, path, audio_settings=None, video_settings=None):
...
+        temp_dir = None
+        if self.REFERENCE_MAPPERS:
+            temp_dir = download_and_rewrite_external_refs(
+                path, mappers=self.REFERENCE_MAPPERS
+            )
+        source = temp_dir or path
+
+        try:
+            # Create partial for reading & compressing subfiles
+            file_converter = partial(
```
rtibbles: Even this code would be better handled by just running every file through the file pipeline (or a reduced version of it).

rtibbles (changes requested):
You seem to have just emulated existing awful code from the downloader (without even deleting much from the module), without taking advantage of any of the affordances of the new pipeline to achieve this in a much more streamlined and consistent way.

If we pass all files through the pipeline, we get free download, file conversion etc - in a completely consistent manner.

ricecooker/utils/pipeline/archive_assets.py
rtibblesbot: Fixed in 5947ff7. External references now go through the pipeline's download stage instead of `make_request`: `download_and_rewrite_external_refs` builds a `DownloadStageHandler` and calls `.execute(url)` per external ref, so archive assets get the same source handling as every other file the pipeline fetches (plain URLs, YouTube, Google Drive, base64, `file://`…) — with retries, caching and content-hashed storage naming for free. `make_request` is gone from the module.

ricecooker/utils/archive_assets.py
rtibblesbot: Fixed in 5947ff7. The hand-rolled fetch-and-write-bytes is gone — the download stage writes each resource to storage under a content-hashed name, and we `shutil.copyfile` it into the archive's `_static/` dir (the pipeline writes to the global storage dir, so the single copy into the archive tree — which is what actually gets zipped — is unavoidable). The `open(source_path)` still on this line is not a copy: it reads a text file (HTML/CSS/JSON) so its references can be rewritten in place, which is the one thing the file pipeline does not do.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Fixed in 5947ff7. `_read_and_compress_archive_file` no longer hand-rolls the compression — it writes the subfile to a temp path and runs it through `ConversionStageHandler.execute`, i.e. the same `VideoCompressionHandler`/`AudioCompressionHandler` used for standalone media. Together with the download-stage change, external and internal archive media now flow through one conversion path. I kept the mp4/webm/mp3 gate so non-media subfiles pass through untouched — routing *every* subfile through the full convert stage would, e.g., reconvert in-archive images to PNG, a behaviour change beyond this PR.

ricecooker/utils/downloader.py
rtibblesbot: `download_static_assets`/`ArchiveDownloader` are a live web *scraper*: they resolve refs against a `base_url` with `urljoin`, recurse into `<a>`/`<iframe>` per `link_policy`, optionally run JS via selenium, and apply js/css middleware hooks. That is a materially different operation from offline archive rewriting, so it cannot collapse into a thin wrapper around `archive_assets`. The part the two genuinely share — CSS/HTML reference *detection* — is already centralized: the regexes live in `references.py` and `downloader` imports them. Migrating its inner extract/rewrite functions fully onto the `references.py` mappers is the #303 refactor over untested legacy scraping paths; I kept it out of this PR to avoid regressing there, and am happy to do it as a focused follow-up.

ricecooker/utils/pipeline/archive_assets.py
```diff
@@ -0,0 +1,202 @@
...
+
+    Holds the shared state (mappers, download stage, assets dir, visited/worklist)
+    so the per-file and per-asset steps read cleanly instead of threading it all
+    through function arguments.
+    """
+
+    def __init__(self, temp_dir, mappers, download_stage, assets_dir):
+        self.temp_dir = temp_dir
+        self.mappers = mappers
+        self.download_stage = download_stage
```
rtibbles: Why just the download stage? What if the file needs to be converted? Why aren't we just reusing most of the pipeline here?

rtibbles (changes requested):
## Rework needed — rearchitect against the design in #233

Design of record: [#233 (this comment)](https://github.com/learningequality/ricecooker/issues/233#issuecomment-4943092980). This is a rearchitecture, not review tweaks — the pipeline-for-all-files direction was right, but the execution built parallel paths and left correctness unverified.

Concrete changes:

- Make `references.py` the *only* detector, a faithful port of kolibri-zip's `fileUtils.js` mappers — not a second one alongside `download_static_assets`.
- Delete `downloader.py` rather than importing its regexes.
- Migrate its live callers to the pipeline: `read` (html_writer, pdf), `get_archive_filename`/thumbnail (chefs).
- Drop the `_get_conversion_stage` module-global.
- Recurse into the running `FilePipeline` via `Handler.parent` instead.
- Route each referenced file through DOWNLOAD → CONVERT — archive media compression now happens there.
- Remove `_read_and_compress_archive_file`.
- Replace the `str(soup)` HTML round-trip with surgical span-replacement.
- Place H5P `content.json` externals under `content/` with a content-relative path.
- `../_static/` will not resolve against `H5P.getPath`.
- Run the mappers for every archive format, including EPUB/KPUB/Bloom.
- Walk all files rather than assuming `index.html` as the entry point.
- That keeps compatibility with #672's denesting and non-root entry points.

The three "reviewer, please confirm" items are correctness questions, not review asks — resolve them against kolibri-zip / `H5PRunner`, don't defer.

ricecooker/utils/pipeline/archive_assets.py
rtibblesbot: Fixed in 3d71585 (with 129e710 and edab23c) — reworked to the design in the issue, not patched.

Each referenced file now goes through the **running `FilePipeline`** (DOWNLOAD → CONVERT), reached via `Handler.parent`, not a bespoke stage. An external ref is downloaded *and* converted — media compressed, incompatible images transcoded — then rewritten to that processed output, in lockstep. The download-stage-only fetch and the `_get_conversion_stage()` singleton are both gone; internal archive media re-enters the *same* running CONVERT stage the same way, so `_read_and_compress_archive_file` no longer hand-rolls anything.

The rest of the rework:
- `references.py` is the sole detector — a faithful port of kolibri-zip's per-file-type mappers (CSS `url()`/`@import`; DOM `src`/stylesheet `href`/`srcset`/inline + `<style>`). `downloader.py` is **deleted** (edab23c), not imported; its live callers — `read` (html_writer, pdf) and the YouTube thumbnail (chefs) — move to the pipeline.
- HTML rewriting is now surgical span-replacement (129e710), not `str(soup)`, which was corrupting third-party apps.
- H5P `content.json` externals land inside `content/` with a content-relative `path` — the only form `H5P.getPath` resolves (`../_static/` would not).
- Every archive format (HTML5/EPUB/KPUB/Bloom) runs the web mappers; H5P adds its content.json mapper. Ref processing walks all files, not just `index.html`.

Correctness items resolved against source, not deferred: kolibri-sandbox's `H5PRunner` keys `contentPaths` by the path *relative to `content/`*, so the asset must live under `content/` — pinned by `test_h5p_json`. Dropped from this PR: a headless JS-render DOWNLOAD handler (the old `read(loadjs=…)` path), which I'll file as a follow-up.

ricecooker/utils/downloader.py
rtibblesbot: Done in edab23c — `downloader.py` is deleted, not wrapped. Its scraping code (`download_static_assets`, `ArchiveDownloader`, `archive_page`, `make_request`) had no in-repo production callers; the only live entry points — `read` (html_writer, pdf) and `get_archive_filename`/thumbnail (chefs) — now go through the pipeline. `read` resolves via the DOWNLOAD stage; the YouTube thumbnail runs the full pipeline (download + webp→png detected from content, not extension). `references.py`'s mappers are the single detector now, and tests/test_downloader.py → tests/test_references.py.

One heads-up: `ricecooker.utils.downloader` was a public util, so downstream chef scripts importing it directly will break on the hard delete. Flagging in case you'd rather ship a thin deprecation shim first.

ricecooker/utils/pipeline/archive_assets.py
rtibblesbot: Superseded by 3d71585. Beyond dropping `make_request`, external refs now go through the **whole** running pipeline (DOWNLOAD → CONVERT), not just the download stage — so they get the same conversion (media compression, image transcode) as every other file. Full write-up on the archive_assets.py:96 thread.

ricecooker/utils/archive_assets.py
rtibblesbot: There is still one `shutil.copyfile` in 3d71585, and it is unavoidable: the pipeline writes each processed resource into the global storage dir under a content hash, and we copy that one file into the archive tree (next to its referencing file) because *that tree* is what gets zipped. No pipeline affordance writes into an arbitrary target dir. The hand-rolled fetch-and-write-bytes and `make_request` are gone.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Done in 3d71585. `_read_and_compress_archive_file` no longer hand-rolls compression and the `_get_conversion_stage()` singleton is gone: internal archive media is routed through the **running** CONVERT stage via `Handler.parent` — the same stage the archive itself is in — and external refs go through DOWNLOAD → CONVERT the same way. Both internal and external media now flow through one pipeline path.

rtibblesbot: One transparent divergence to flag: `_read_and_compress_archive_file` still exists as a *thin* delegator (writes the subfile out, re-enters the running CONVERT stage, returns the result) rather than being removed in favour of a purely reference-driven walk. Reason: the whole-archive `create_predictable_zip` pass compresses **every** media subfile, whereas a reference-only walk would compress just the statically-referenced ones. That is the same "static reachability under-approximates, so don't prune" argument you made for keeping files — media pulled in dynamically (JS, H5P libraries) is reachable at runtime but invisible to a static ref scan, so compressing only referenced media would silently skip it. So the whole-archive compression stays, but now goes through the running pipeline via `Handler.parent` with no bespoke logic or global. Happy to collapse it further if you'd prefer the reference-only semantics despite that gap.

ricecooker/utils/pipeline/archive_assets.py
```diff
@@ -0,0 +1,221 @@
+"""Download external resources referenced by an archive and rewrite the refs.
```
rtibbles: Difficult for me to review this at the moment, because the code comments are so extraordinarily long, doing a reasonable review is challenging.

rtibbles (changes requested):
I have no idea why it was deemed necessary to write multiple essays for module level doc strings, class, and method doc strings.

You should be exercising the /terse-writing discipline for this technical writing - not writing extraordinarily verbose code comments.

ricecooker/utils/pipeline/archive_assets.py
rtibblesbot: Done in fd6366c. Ran the /terse-writing discipline over the docstrings and comments this PR added. The archive_assets.py module docstring drops from ~40 lines to 16; its class/method docstrings and the per-file/per-asset comments are cut to the load-bearing facts. Same pass over references.py, convert.py, transfer.py and chefs.py. No behavior change — lint clean, tests/test_references.py + tests/pipeline/test_convert.py green (36 passed).

ricecooker/utils/pipeline/archive_assets.py
```diff
@@ -0,0 +1,170 @@
...
+    downloaded asset a mapper handles (e.g. CSS referencing fonts) is re-scanned,
+    with a visited-URL guard. A failed download leaves the reference unrewritten.
+
+    ``pipeline`` defaults to a fresh ``FilePipeline`` (imported lazily to avoid an
+    import cycle); a caller inside a running pipeline passes its own to reuse the
+    same stages, config and caches.
+
+    Returns the extracted temp directory. **The caller owns cleanup.**
+    """
+    if pipeline is None:
```
rtibbles: We should remove this - just make pipeline a required argument - its only caller already passes the pipeline in, so this optionality is just extra confusion.

ricecooker/utils/pipeline/convert.py
```diff
@@ -193,29 +202,60 @@ def FILE_TYPE(self) -> str:
...
-        # create_predictable_zip will iterate over subfiles, call file_converter
-        processed_zip_path = create_predictable_zip(
-            path, file_converter=file_converter if config.COMPRESS else None
-        )
+        # Download external references and rewrite them to local copies before
+        # create_predictable_zip seals the archive. Handlers with REFERENCE_MAPPERS
+        # get a fresh temp dir; the rest zip the path as-is.
+        temp_dir = None
+        if self.REFERENCE_MAPPERS:
+            temp_dir = download_and_rewrite_external_refs(
```
rtibbles: I don't understand why we are returning a directory of files here - this means that all of our code below has to operate _either_ on a zip file, or on a temporary directory path.

We should own creating the temp dir inside `handle_file`, then apply the _ArchiveProcessor to it. The function itself is a pointless wrapper. The ArchiveProcessor should be the public API.

ricecooker/utils/pipeline/convert.py
rtibbles: The ArchiveProcessor should be doing two passes - one a reference led walk to ensure all referenced assets are reachable, and another to ensure all assets that can be compressed are compressed.

Basically, I think we should be consolidating as much as we can onto the archive processor, which just takes a directory as input, processes all references, and compresses all media files that it can - then hand it off to create_predictable_zip which just concerns itself with compressing the directory into a new archive.

rtibbles (changes requested):
This is much better, much more readable - I still think we are still deferring too much to the previous architecture, which is making things more convoluted than they ought to be.

The basic proposal is, the archive handler handles:
* Unzipping the archive into a temporary directory
* Passing to the ArchiveProcessor (perhaps it's actually just a DirectoryProcessor?) to handle all recursive pipeline operations (full pipeline pass + scoped conversion pass)
* Running create predictable zip to recreate the archive file

ricecooker/utils/pipeline/archive_assets.py
rtibblesbot: Done in 804d07a. `pipeline` is now a required positional argument on `ArchiveProcessor` (the wrapper that carried the `pipeline=None` default and the fresh-`FilePipeline` fallback is gone). The only caller — `handle_file` — passes `self._running_pipeline()` in.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Done in 804d07a. `download_and_rewrite_external_refs` is gone. `handle_file` now owns the temp dir: it unzips the archive into a `mkdtemp()`, applies `ArchiveProcessor` to that directory, then hands the directory to `create_predictable_zip`. Nothing downstream branches on zip-vs-directory anymore, and `ArchiveProcessor` is the public API.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Done in 804d07a. `ArchiveProcessor.process()` now runs two passes over the directory: (1) a reference-led walk that fetches every external ref through the running pipeline so referenced assets are reachable, then (2) a scoped conversion pass that recompresses every media file it can, in place. `_read_and_compress_archive_file` and the `file_converter` callback are gone; `create_predictable_zip(temp_dir)` now only zips the finished directory into a new archive.

ricecooker/utils/pipeline/archive_assets.py
```diff
@@ -0,0 +1,203 @@
...
+    Operates on ``directory`` in place, reusing the running ``pipeline`` so an
+    external reference is downloaded and converted through the same stages, config
+    and caches as every other file. Detection is delegated to ``mappers``; the
+    first that handles a file scans it.
+    """
+
+    # Media the pipeline can recompress in place without changing the container
+    # (and thus without invalidating references to it); everything else is passed
+    # through untouched, since a changed extension would break its references.
+    MEDIA_EXTENSIONS = {file_formats.MP4, file_formats.WEBM, file_formats.MP3}
```
rtibbles: Can we not infer this, rather than hardcoding it? We have the explicit metadata for this, so we can avoid creating a separate source of truth here.

ricecooker/utils/archive_assets.py
```diff
@@ -0,0 +1,203 @@
+"""Process an extracted archive in place: download external refs, compress media.
```
rtibbles, line 1: As this doesn't _actually_ define handlers or a stage in the pipeline, I think I would rather this still live under the general utils module, rather than specifically in the pipeline (this is also suggested by the fact that the mappers/references live in the utils module directly too).

ricecooker/utils/archive_assets.py
```diff
@@ -0,0 +1,203 @@
...
+
+from le_utils.constants import file_formats
+
+from ricecooker import config
+from ricecooker.config import LOGGER
+from ricecooker.utils.pipeline.exceptions import ExpectedFileException
+from ricecooker.utils.pipeline.exceptions import InvalidFileException
+from ricecooker.utils.references import DEFAULT_MAPPERS
+from ricecooker.utils.references import is_external_url
+from ricecooker.utils.utils import extract_path_ext
```
rtibbles, line 28: utils.utils is such a stupid module path - we should file a follow up issue to clean this up and make things more sensible.

ricecooker/utils/pipeline/archive_assets.py
```diff
@@ -0,0 +1,203 @@
...
+        self.audio_settings = audio_settings or {}
+        self.video_settings = video_settings or {}
+        # Fetched URL -> pipeline output path (None on failure): fetch each URL
+        # once and terminate cycles in downloaded CSS.
+        self.visited = {}
+
+    def process(self):
+        """Download external refs, then compress media, in place."""
+        self._download_external_refs()
+        if config.COMPRESS:
```
rtibbles: In this PR: https://github.com/learningequality/ricecooker/pull/673

We remove the global COMPRESS flag. Could you bring those changes into this PR as a discrete commit, so we can avoid making more churn to clean this up? We can then close that PR in preference to this one.

ricecooker/utils/pipeline/convert.py
```diff
@@ -193,29 +201,48 @@ def FILE_TYPE(self) -> str:
...
 
+    def _running_pipeline(self):
+        """The FilePipeline this handler is executing inside, or ``None``.
+
+        Archive handlers are grandchildren of the pipeline (via the CONVERT
+        stage). Reusing it means an external reference is downloaded and converted
+        through the same stages, config and caches as every other file.
+        """
+        # Imported lazily: ricecooker.utils.pipeline imports this module.
+        from ricecooker.utils.pipeline import FilePipeline
```
rtibbles: When would it not be an instance of FilePipeline? It's either a pipeline or its None? I think we can add this resolution to the root handler method "get pipeline" which can then just recurse back up until it finds it, without this awkward inline import.

Still confused why this wouldn't be defined too.

tests/pipeline/test_convert.py
```diff
@@ -247,3 +272,288 @@ def test_invalid_zip(self):
                 f.write(b"not a zip file")
             with pytest.raises(InvalidFileException, match="(?i)zip"):
                 KPUBConversionHandler().validate_archive(path)
+
+
+class _FakePipeline:
```
rtibbles: Why are we faking a pipeline here? This goes against our integration testing focus. We should mock at the external boundaries (requests etc) but otherwise exercise real code paths to ensure that our code actually all works together. As soon as we mock an internal API, we can change internal APIs without breaking tests.

rtibbles (changes requested):
It keeps on getting better - still some more cleanup to do, but we're on a clear trajectory towards success.

When you have done the clean up here, please make sure to reorder the commit history - it's currently a very long flip flopping mess, so we want to turn it into a sensible description of the current state of work. Also update the PR body if it is out of date.

ricecooker/utils/pipeline/archive_assets.py
rtibblesbot: Done (in `d336518`). Dropped the hardcoded `MEDIA_EXTENSIONS` set — `media_extensions` now reads the recompressable containers straight off the CONVERT stage's own media handlers (`VideoCompressionHandler.SUPPORTED_VIDEO_EXTS` / `AudioCompressionHandler.SUPPORTED_AUDIO_EXTS`). Single source of truth: anything the stage would re-encode to a new extension (e.g. gif→png), breaking references to it, declares no such set and is skipped.

ricecooker/utils/archive_assets.py
rtibblesbot: Done — moved `ricecooker/utils/pipeline/archive_assets.py` → `ricecooker/utils/archive_assets.py`, alongside `references.py`. It defines neither a handler nor a stage, so it no longer lives in the pipeline package. `convert.py` imports it lazily inside `handle_file` (the module depends on the pipeline's exceptions, so a top-level import would be circular).

ricecooker/utils/archive_assets.py
rtibblesbot: Filed as #692 (https://github.com/learningequality/ricecooker/issues/692) — rename the `ricecooker.utils.utils` module. Kept out of this PR to avoid unrelated churn across every import site.

ricecooker/utils/pipeline/archive_assets.py
rtibblesbot: Done in `c57a9da`, as a discrete commit. `config.COMPRESS` is removed; `SushiChef.run` derives explicit ffmpeg settings from `--compress` and threads them through the pipeline's new `default_context`, so handlers compress only when settings are provided (standalone and in-archive media alike). #673 can be closed in preference to this.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Done in `3777f06`. Added `Handler.get_pipeline()` on the base handler, which walks up the `parent` chain to the root (the pipeline) — no isinstance guard and no inline import. The archive handler just calls `self.get_pipeline()`. In production the root is always the running `FilePipeline`; a handler exercised outside a pipeline resolves to itself, and the integration tests now always build a real pipeline (see the `test_convert.py` thread).

tests/pipeline/test_convert.py
rtibblesbot: Done in `3777f06`. Dropped `_FakePipeline`. The `ArchiveProcessor` tests now run a real `FilePipeline` and mock only the external boundary — `config.DOWNLOAD_SESSION` — so each external ref flows through the true download → convert path (an unmapped URL raises like a failed request, exercising the leave-unrewritten branch). The end-to-end handler test likewise drives `HTMLZipFile.process_file()` with only the download session faked.

rtibbles (approved):
Good changes, satisfactory test coverage.

ricecooker/utils/downloader.py
jaltekruse: @rtibbles I understand this deletion likely was assumed pretty low impact given the current volume of new sushichef script development. I was using this achive_page method in my script for importing pretext books. I am planning on just refactoring my script to get it working with the current tip of main, but it looks like both you and claude talked about keeping these APIs around with thin wrappers around the refactored code. So I thought I might call it out as a potential bit of work to restore these APIs if there are other scripts that might be affected by this file deletion.

PR #695, opened 2026-07-21
Convert article documents to KPUB via pandoc; sanitize all KPUB CSS

### LINKED ISSUE #685: Convert article documents to KPUB in the pipeline (retire the dead microwave service)
❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

## Overview

Retire the defunct `microwave` service and give article-style documents a first-class path to **KPUB** (self-contained HTML article). A chef references a source document and it becomes a validated KPUB automatically:

```python
ContentNode(uri="article.docx")   # → validated KPUB in the channel
```

**Complexity:** Medium
**Target branch:** main

### Context

- `microwave` (`http://35.185.105.222:8989/unoconv/pdf`) is dead.
- It is referenced only in `docs/examples/document_conversion.ipynb` — no runtime code uses it.
- It documented a manual workflow: convert an office doc to PDF out-of-band, then add the PDF.
- KPUB is already supported end-to-end (`KPUBConversionHandler` + `KPUBMetadataExtractor`); no path exists to *produce* one from a source document.
- The CONVERT stage rewrites the working file's extension, so a handler emitting `.kpub` feeds the existing `KPUBMetadataExtractor` — net new code is one handler plus the sanitizer.
- Kolibri's KPUB renderer strips all CSS at runtime; ricecooker's validator only forbids external `.css`.
- Selector-bearing `<style>` can leak into Kolibri chrome; element-scoped `style=` is safe when bounded to an allowlist.
- The Studio rich text editor (TipTap v3) emits exactly one inline style — `text-align` — on `<p>`, `<h1>`–`<h3>`, `<small>`, ``; every other format is a semantic tag.

### The Change

- Add a CONVERT-stage handler converting `.docx`, `.odt`, `.rtf`, `.md`, `.markdown` → KPUB via `pandoc`.
- Emit **semantic-only** HTML: images as sibling files (`--extract-media`), math as presentation MathML.
- Tighten `KPUBConversionHandler` into a sanitizer applied to all KPUB (converted and hand-authored).
- Sanitizer: strip every `<style>` block.
- Sanitizer: on `style=`, keep only allowlisted properties, drop the rest.
- Sanitizer: strip offending CSS rather than reject the file; keep the existing `<script>`/`.js`/`.css` bans.
- CSS `style=` **allowlist v1 = `{text-align, color, background-color}`**, a ricecooker-local constant. `text-align` from the Studio RTE; `color`/`background-color` for source-doc text and highlight colours. Widening further requires extending the RTE first.
- Change `GoogleDriveHandler` to export Google Docs as `.docx` (currently PDF); Google Slides stay PDF.
- Rewrite `docs/examples/document_conversion.ipynb` around the new workflow; remove the dead service and IP.
- Document `pandoc` as a system dependency (alongside ffmpeg/poppler) and install it in CI on all supported OSes.
- Render math as MathML (MathJax needs no LaTeX present); a LaTeX `<annotation>` for source preservation is a verify-then-implement enhancement, not a blocker.

### Out of Scope

- Presentations (`.pptx/.ppt`) and spreadsheets (`.xlsx/.xls`) — not article content.
- Legacy `.doc` — pandoc can't read the binary format.
- `.epub` (native EPUB) and `.html`/zipped HTML (HTML5) — not claimed by the new handler.
- Widening the allowlist beyond `{text-align, color, background-color}`.
- The coupled Kolibri renderer change to honor the allowlist instead of stripping all CSS — tracked as a separate Kolibri issue.
- Any hosted or self-hosted conversion service.

### Acceptance Criteria

- [ ] `ContentNode(uri="article.docx")` (and `.odt`, `.rtf`, `.md`, `.markdown`) produces a valid KPUB through the pipeline with no manual pre-conversion.
- [ ] Converted output has `index.html` with a non-empty body and no `<script>`/`.js`/`.css`/`<style>`.
- [ ] Converted images land under `media/` and are referenced by relative path.
- [ ] Math in a source document appears as presentation MathML (`<math>`) in the output.
- [ ] `KPUBConversionHandler` strips all `<style>` blocks from any KPUB (converted and hand-authored).
- [ ] `style=` with `text-align`, `color`, or `background-color` survives sanitization.
- [ ] A non-allowlisted `style=` property (e.g. `position`) is dropped; a mixed attribute keeps only the allowlisted properties.
- [ ] A hand-authored KPUB with disallowed CSS still ingests, with a log line naming what was removed.
- [ ] Google Docs downloaded via `GoogleDriveHandler` export as `.docx` and convert to KPUB.
- [ ] Google Slides still export as PDF.
- [ ] `pandoc` is documented as a system dependency and installed in CI on all supported OSes.
- [ ] A clear, actionable error is raised when pandoc is absent.
- [ ] `docs/examples/document_conversion.ipynb` no longer references `microwave` or the dead IP.
- [ ] Tests cover each source format converting to a valid KPUB.
- [ ] Tests cover the sanitizer keep/drop cases.
- [ ] Tests cover the pandoc-missing error and the GDrive export-format change.

### References

- `ricecooker/utils/pipeline/convert.py` — `KPUBConversionHandler`, `ConversionStageHandler`
- `ricecooker/utils/pipeline/transfer.py` — `GoogleDriveHandler.GOOGLE_WORKSPACE_FORMATS`
- Studio RTE (allowlist source of truth): `contentcuration/.../frontend/shared/views/TipTapEditor/`

## AI usage

Used Claude Code (Opus) to draft this issue and map the codebase from a live brainstorming dialogue that I directed. Codebase claims (microwave is docs-only; the RTE emits only `text-align`) were verified against the ricecooker and studio repos.

rtibblesbot:
## Summary

Referencing an article-style source document from a chef — `ContentNode(uri="article.docx")` (also `.odt`, `.rtf`, `.md`, `.markdown`) — now produces a validated KPUB automatically, with no manual pre-conversion through the defunct `microwave` service. A new CONVERT-stage handler shells out to `pandoc` to render semantic-only HTML (images extracted to `media/`, math as presentation MathML) and reseals it as `.kpub`, which the existing metadata extractor picks up unchanged. Every KPUB — converted or hand-authored — is now run through a surgical sanitizer that strips `<style>` blocks, `<script>` elements, and the IE conditional comments some pandoc templates wrap an html5shiv `<script>` in, and filters inline `style=` down to an allowlist (`text-align`, `color`, `background-color`), so disallowed CSS or scripts cannot leak into Kolibri chrome. The HTML reference rewriter also stops resolving `<iframe src>` — a navigation link like `<a href>`, not an offline resource — so a KPUB that iframes remote HTML no longer downloads that page unsanitized. Google Docs export as `.docx` so Drive documents flow through the same path; Google Slides stay PDF. `pandoc` is documented as a system dependency (alongside ffmpeg/poppler) and installed in CI on every supported OS, and the `document_conversion.ipynb` example is rewritten around the new workflow.

## References

Fixes #685.

## Reviewer guidance

- `ricecooker/utils/pipeline/convert.py` — `PandocMissingError` is deliberately excluded from `DocumentConversionHandler.HANDLED_EXCEPTIONS`, so a missing `pandoc` binary raises loudly instead of being swallowed as a per-file `ExpectedFileException`; confirm that is the intended failure mode for a missing system dependency.
- `ricecooker/utils/references.py` — `sanitize_style_css` strips `<style>` blocks with a regex pass, then `_HTMLStyleSanitizer` splices inline `style=` attributes down to the allowlist rather than reparsing; check the forward-only cursor handles adjacent/repeated `style=` attributes without corrupting neighbouring bytes. `strip_scripts` removes IE conditional comments before `<script>` blocks (the html5shiv script lives inside one) — confirm that ordering.
- `ricecooker/utils/pipeline/transfer.py:285` — confirm `presentation` (Google Slides) still maps to PDF.
- `ricecooker/utils/pipeline/transfer.py:283` — confirm Google Docs now export as `.docx` so they route into `DocumentConversionHandler`.

## AI usage

Used Claude Code (Opus) to implement the plan test-first. Verified with the full pipeline and references test suites (with `pandoc` installed locally) and `uvx prek run --all-files`.

---

</details>

<table><tr><td><a href="https://tibbs.io/status">@rtibblesbot</a></td><td><h3>🟡 Waiting for feedback</h3><em>Last updated: 2026-07-22 22:43 UTC</em></td></tr></table>


ricecooker/utils/pipeline/convert.py
```diff
@@ -603,6 +647,47 @@ def handle_file(self, path, language=None, subtitle_format=None):
...
+                    text=True,
+                    check=True,
+                )
+            except subprocess.CalledProcessError as e:
+                raise PandocConversionError(
+                    f"pandoc failed to convert {path}: {e.stderr}"
+                )
+            # pandoc --extract-media localizes embedded media only; unlike an
+            # uploaded KPUB, remote <img> refs are not downloaded (out of scope).
+            sanitize_kpub_directory(temp_dir)
```
rtibbles, line 689: Could we just run it through the KPUB extractor?

ricecooker/utils/pipeline/convert.py
```diff
@@ -52,6 +56,43 @@
...
+
+class PandocMissingError(Exception):
+    """Raised when the pandoc system binary is required but not installed."""
+
+
+class PandocConversionError(Exception):
+    """Raised when pandoc fails to convert a source document."""
+
+
+def sanitize_kpub_directory(temp_dir):
```
rtibbles, line 71: Could this not be handled as a reference mapper? Feels like that is the more natural place for this to happen.

rtibbles (commented):
This is looking pretty good, just a couple of questions about approach.

ricecooker/utils/pipeline/convert.py
rtibblesbot: The KPUB extractor can't do the sanitizing — `KPUBMetadataExtractor` is a bare `MetadataExtractor` subclass that only maps the `.kpub` extension to `format_presets.KPUB_ZIP`. The EXTRACT_METADATA stage reads metadata; it never unpacks, modifies, or reseals the archive, so it can't strip CSS/scripts.

And CONVERT is `FirstHandlerOnly` (`ConversionStageHandler(StageHandler)` → `StageHandler(FirstHandlerOnly)`), so once this handler consumes the `.docx` and emits a `.kpub`, that file doesn't re-enter `KPUBConversionHandler` — the sanitization has to happen here, in the same pass that seals the archive.

It isn't duplicated logic though: both this handler and `KPUBConversionHandler.process_extracted` call the same `sanitize_kpub_directory` helper, so hand-authored and pandoc-converted KPUB go through identical sanitization. Happy to reshape if you had a different chokepoint in mind.

ricecooker/utils/pipeline/convert.py
rtibblesbot: I looked at that — the `ReferenceMapper` contract is URL-centric: `map(content, fn: url -> url) -> (content, urls)`, with `extract`/`rewrite` on top, all for finding and rewriting external references so `ArchiveProcessor` can download them locally. Sanitization removes/filters content (drop non-allowlisted declarations, strip `<style>`/`<script>`) — there's no URL to map, so it doesn't fit that interface.

Two other blockers:
- `DEFAULT_MAPPERS` run on every archive format via `ArchiveProcessor`. Style/script stripping is KPUB-specific policy — we don't want it applied to H5P/Bloom/HTML5.
- `DocumentConversionHandler` doesn't run `ArchiveProcessor` at all (it's not an `ArchiveProcessingBaseHandler`), so a mapper would never fire for pandoc-converted docs. The sanitizer has to be reachable outside the mapper pass, which is why it's a shared helper called from the KPUB-producing handlers.

That said, the reusable string primitives already live in `references.py` next to the mappers and share the same surgical `_apply_edits` machinery (`sanitize_style_css`, `strip_scripts`) — only the KPUB policy (the allowlist + file I/O) sits in `convert.py` with the KPUB handlers. Happy to revisit if you'd still prefer it modeled as a mapper subclass.

ricecooker/utils/references.py
```diff
@@ -223,6 +233,101 @@ def _map_html_urls(html: str, fn: Callable[[str], str]) -> Tuple[str, List[str]]
...
+            kept.append(decl.strip())
+        else:
+            dropped.append(prop)
+    return "; ".join(kept), dropped
+
+
+_STYLE_BLOCK_RE = re.compile(r"<style\b[^>]*>.*?</style\s*>", re.IGNORECASE | re.DOTALL)
+
+
+class _HTMLStyleSanitizer(_SurgicalHTMLParser):
```
rtibbles, line 258: Ohh... I know what my concern was now. If we don't do sanitization at the same time as we're doing remote resource resolution, we could end up downloading a bunch of stuff that we don't need. This is why sanitization can't be a final step, it has to happen at the same time as the reference resolution, because the sanitization has to be applied to the reference resolution also.

rtibbles (changes requested):
Code has been put into references.py to kind of pretend like it's integrated, but it doesn't cover the specific issue of not sanitizing simultaneously to resolving downloadable references.

ricecooker/utils/references.py
rtibblesbot: Agreed — that ordering was the bug. Fixed in e21b13e by moving `process_extracted` (the sanitization hook) to run *before* `ArchiveProcessor.process()` in `ArchiveProcessingBaseHandler.handle_file`, rather than after it.

So the sequence is now: extract → sanitize → resolve references → reseal. Because the `<style>` blocks and non-allowlisted `style=` declarations are stripped first, the `ArchiveProcessor` reference walk only ever sees — and only ever downloads for — content that survives sanitization. A `url()` inside a `<style>` block, or a `background-image:url(...)` that the allowlist drops, no longer triggers a download that would then be orphaned.

This keeps sanitization as a content-filtering pass distinct from the URL-mapping `ReferenceMapper` interface (per the other thread), while still guaranteeing it is applied to the input of reference resolution.

Added a full-pipeline regression test (`test_stripped_refs_are_not_downloaded`) that fails under the old ordering: it asserts a resource referenced only from a stripped `<style>` block and from a dropped `style=` property is never fetched, while a plain `` still is.

rtibbles: Hrm, are we blocking <a> tags? Because otherwise this would allow downloading an HTML file linked via an <a> tag, right? And then that hasn't been sanitized.

rtibblesbot: Yes — `<a href>` (and `<iframe src>`) are already excluded from reference resolution, so nothing is ever downloaded through them.

`_HTMLReferenceRewriter._handle_tag` (references.py:167-182) only records *offline resource* references: `src` on any element, stylesheet `<link href>` (via `_is_stylesheet_attrs`), `srcset`, and inline/block CSS `url()`/`@import`. Plain `href` is recorded **only** for stylesheet `<link>`s — never for `<a>`. Navigation links are deliberately skipped; the rewriter docstring calls this out explicitly (references.py:141-143: "Navigation references (`<a href>`, `<iframe src>`) are excluded: they are page links, not offline resources").

So the failure mode you describe — an HTML file pulled in via an `<a>` link and then sitting unsanitized in the archive — cannot occur: `ArchiveProcessor` never sees that URL, so it never fetches it. The only things it downloads are images/media/stylesheets, not documents.

This is locked in by tests:
- `test_extract_html_urls` (tests/test_references.py:70-80) asserts a navigation/`canonical` `href` is **not** among the extracted URLs.
- `test_rewrite_is_surgical` (tests/test_references.py:97-131) asserts an `<a href>` is left byte-for-byte untouched during rewriting.

ricecooker/utils/pipeline/convert.py
```diff
@@ -208,14 +255,21 @@ def handle_file(self, path, audio_settings=None, video_settings=None):
...
         with tempfile.TemporaryDirectory() as temp_dir:
             with zipfile.ZipFile(path) as zf:
                 zf.extractall(temp_dir)
 
+            # Sanitize BEFORE reference resolution, not after. A <style> block or
+            # a non-allowlisted style= declaration can carry a url(...) reference;
+            # resolving references first would download those resources only for the
+            # sanitizer to strip the content that referenced them, leaving orphaned
+            # assets in the archive.
+            self.process_extracted(temp_dir)
```
rtibbles: I think we should call this `pre_process` just to make it clear this is happening before other processing steps.

rtibbles (changes requested):
Just one small naming tweak, and this should be good to go. Make the correction, then tidy up the commit history.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Done — renamed the hook (and its call site + the `KPUBConversionHandler` override) to `pre_process`. Since I was tidying the history anyway, this is folded into the `Sanitize all KPUB…` commit (79e33f5) so the hook reads as `pre_process` throughout, rather than being renamed in a later commit.

rtibbles (changes requested):
Rebase on main please.

tests/cassettes/test_gdrive_doc.yaml
```diff
@@ -121,7 +121,7 @@ interactions:
       x-goog-api-client:
       - gdcl/2.161.0 gl-python/3.10.10 cred-type/sa
     method: GET
-    uri: https://www.googleapis.com/drive/v3/files/1T9dM1gbOc_aOXwvs8H1bkqSfJhcnLG11yZQ9RPEJI6Y/export?mimeType=application%2Fpdf&alt=media
+    uri: https://www.googleapis.com/drive/v3/files/1T9dM1gbOc_aOXwvs8H1bkqSfJhcnLG11yZQ9RPEJI6Y/export?mimeType=application%2Fvnd.openxmlformats-officedocument.wordprocessingml.document&alt=media
```
rtibbles, line 124: How can it be that the output format has changed, but the actual binary response is the same? This isn't right.

rtibbles (changes requested):
Dodgy fixture.

tests/cassettes/test_gdrive_doc.yaml
rtibblesbot: Good catch — the recorded response body was stale PDF bytes from before the export-mime change, so the cassette claimed a `.docx` request but replayed a PDF. Fixed: the export interaction now replays a real `.docx` (an OOXML zip — `PK\x03\x04` magic bytes) generated from the same "This is a sample document" content, with `Content-Length` corrected to match (10372) and `Content-Type` already the OOXML mime. The test asserts the derived filename extension, but the cassette is now faithful to the request it records.

Also rebased on `main` as requested; both changes are in the force-pushed branch.

ricecooker/utils/references.py
rtibblesbot: Correction / follow-up: I re-checked this against the actual code rather than the docstring, and `<iframe src>` was **not** in fact excluded — the `src`-on-any-element branch in `_HTMLReferenceRewriter._handle_tag` resolved it, so a KPUB linking remote HTML via an `<iframe>` would have downloaded that page unsanitized (exactly your concern). Only `<a href>` was genuinely excluded.

Fixed in 49426b8: iframe `src` is now skipped, matching the module docstring and kolibri-zip's `DOMMapper`. Added an `<iframe>`/`<a>` case to `test_extract_html_urls` asserting neither is fetched. Sorry for the earlier overclaim.

rtibbles (approved):
Good work!

Issue #691, opened 2026-07-11
Headless page-archiving: wrap single-file-cli, explode through the pipeline


rtibbles:
<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

❌ **This issue is not open for contribution. Visit <a href="https://learningequality.org/contributing-to-our-open-code-base/" target="_blank">Contributing guidelines</a>** to learn about the contributing process and how to find suitable issues.

<img height="20px" src="https://i.imgur.com/c7hUeb5.jpeg">

## Overview

Re-add headless page-archiving (render a JS/SPA page, follow links, snapshot to an HTML5 zip) by wrapping single-file-cli for capture and reusing the #233 pipeline for localization and conversion. This restores a capability the #233 rework removes.

**Complexity:** High
**Target branch:** `main`

### Context

- #233 deletes `downloader.py`, which held the only headless page-archiving path: `read(loadjs=…)`, `archive_page`, `ArchiveDownloader`, `download_in_parallel`, `link_policy`.
- It was dropped rather than ported because no currently maintained content integration scripts use it.

### The Change

- Wrap **single-file-cli** for capture — it renders JS via Chromium over CDP.
- It crawls with `--crawl-links` / `--crawl-max-depth` / `--crawl-inner-links-only` / `--crawl-rewrite-rule`, replacing the old `link_policy`.
- Its output is one self-contained HTML per page, assets inlined as `data:` URIs.
- Feed that HTML through the #233 reference-driven archive processing.
- Extend that processing to treat `data:` URIs as localizable references.
- The existing Base64 DOWNLOAD handler decodes each `data:` URI to a file.
- CONVERT then compresses media and converts incompatible images.
- Refs are rewritten to the exploded files, so the result is a real HTML5 zip rather than one inlined file.
- Expose it as a dedicated, non-default DOWNLOAD-stage handler that page-archiving scripts enable explicitly.
- The source marks render intent; nothing is auto-detected.

### Out of Scope

- Owning a browser engine or building a Playwright/Puppeteer renderer.
- WARC capture or replay.
- Sites needing authentication or heavy anti-bot measures beyond single-file-cli's reach.

### Acceptance Criteria

- [ ] A page-archiving script can render a JS/SPA page via single-file-cli through the pipeline and produce an HTML5 zip.
- [ ] Crawl depth and link scope are configurable (link-following parity with the old `link_policy`).
- [ ] `data:` URIs in captured HTML are exploded into separate files.
[...]

PR #714, opened 2026-07-24
Render HTML5-zip and KPUB thumbnails via headless Chromium

### LINKED ISSUE #335: Better HTML5Zip thumbnail generation
The studio code has a much better way to generate thumbnails from zip files with web content:
load the index.html in a headless browser and take a screenshot. See 
https://github.com/learningequality/studio/blob/f57e27a8898faecd99f4d4d97109cb50eab7e74a/contentcuration/contentcuration/utils/files.py#L342-L407

We should use this approach for [`create_image_from_zip`](https://github.com/learningequality/pressurecooker/blob/master/pressurecooker/images.py#L88-L120), but not implemented with PhantomJS.

Waiting on https://github.com/learningequality/ricecooker/issues/176 to be fixed first so we can use the same tooling.

rtibblesbot:
## Summary
HTML5-zip thumbnails were derived by cropping the largest embedded raster image, which produces nothing usable for canvas/SVG/text apps and picks unrepresentative sprites or backgrounds otherwise. This renders the archive's resolved HTML entry point in a headless Chromium and screenshots the viewport, smart-cropping it to the thumbnail spec. KPUB articles get the same rendered-screenshot path. The render shells out to an externally installed Chromium binary, so the core install is unaffected and thumbnail generation falls back to the biggest-image heuristic both when Chromium is absent and when a render fails. The existing HTML entry-point resolver (previously private in `convert.py`) is promoted to `utils/zip.py` so both the pipeline and the screenshot path share one implementation instead of duplicating it.

## References
Closes #335. Related: #691 (single-file render), #708 (PhantomJS removal).

Acceptance criteria (issue #335):
- [x] No-raster-image HTML5 zip produces a rendered screenshot
- [x] KPUB archive produces a rendered screenshot
- [x] Entry point resolved via detection, not hardcoded `index.html`
- [x] Screenshot smart-cropped and scaled to the thumbnail spec
- [x] No browser dependency added beyond Chromium
- [x] Chromium absent falls back to the biggest-image heuristic
- [x] Core install unaffected when Chromium is absent
- [x] Render path tested against a no-image zip
- [x] KPUB path tested
- [x] Chromium-absent fallback tested

## Reviewer guidance
- `ricecooker/classes/files.py:568` — both `ChromiumUnavailableError` (Chromium missing) and `ThumbnailGenerationError` (render failed) fall back to the biggest-image heuristic; render failures are logged as a warning and swallowed rather than marking the file failed. Confirm that silently degrading a failed render to biggest-image is the intended behaviour.
- `ricecooker/utils/images.py` — the screenshot invokes Chromium with `--headless=new` and `--no-sandbox`; confirm these flags suit the CI/container environments where thumbnails are generated.
- `ricecooker/classes/nodes.py:1099` — the KPUB branch sits after pdf/epub in the `if/elif`; confirm KPUB and PDF/EPUB presets are mutually exclusive on a `DocumentNode`.

## AI usage
Used Claude Code to implement the render-and-screenshot thumbnail path and its tests. Verified with the thumbnail and zip test suites and ruff.

---

</details>

## Deviations from the issue spec
- The spec only requires falling back to the biggest-image heuristic when Chromium is absent; this branch also falls back when the render itself fails (e.g. Chromium timeout/crash in headless CI), logging a warning instead of failing the file, to avoid regressing prior thumbnail behaviour.

<table><tr><td><a href="https://tibbs.io/status">@rtibblesbot</a></td><td><h3>🟡 Waiting for feedback</h3><em>Last updated: 2026-07-24 19:30 UTC</em></td></tr></table>


ricecooker/utils/zip.py
```diff
@@ -8,6 +8,34 @@ def _read_file(path):
         return f.read()
 
 
+def find_html_entrypoint(names):
```
rtibbles: Don't we _already_ have logic for discovering the HTML entrypoint? Why have we created an entirely new function here?

rtibbles (changes requested):
Haven't manually tested yet - but one concern about code duplication.

ricecooker/utils/zip.py
rtibblesbot: Good catch — you're right. `convert.py` already had `_find_entry_html`/`_find_common_root` (the Studio `findFirstHtml`/`findCommonRoot` port from #672), and this new function duplicated it.

Fixed in 1ecc593: promoted that canonical logic into the low-level `utils/zip.py` (stdlib-only, correct dependency direction) as `find_html_entrypoint`/`find_common_root`, and `convert.py` now imports it instead of keeping its private copies. Both the screenshot path and the pipeline HTML5/KPUB validation now share one implementation. Behavior is the tested `convert.py` version, unchanged; its unit tests moved to `tests/test_zip.py` alongside the function.

ricecooker/utils/images.py
```diff
@@ -106,6 +111,83 @@ def create_image_from_zip(htmlfile, fpath_out, crop="smart"):
...
+                "--no-sandbox",
+                "--disable-gpu",
+                "--hide-scrollbars",
+                "--force-device-scale-factor=1",
+                "--window-size=1600,900",
+                "--user-data-dir={}".format(profile_dir),
+                "--screenshot={}".format(shot_path),
+                url,
+            ]
+            subprocess.run(cmd, capture_output=True, timeout=60, check=True)
```
rtibbles, line 171: Add `--virtual-time-budget=<ms>` to `cmd` above so Chromium settles timers/network before this capture (see review summary).

rtibbles (changes requested):
Verified against a real html5 zip (`tests/testcontent/samples/sample_html.zip`): the thumbnail comes out fully blank white.

The screenshot is taken immediately with no render wait, so any JS-driven app — the common case — paints after capture. Confirmed by running the exact flag set: blank as-is, correct cover page once `--virtual-time-budget=8000` is added.

Worse, it's a *successful* blank, not an error, so the biggest-image fallback (0ca89aa) never fires — and here that fallback (`IMAGES/CoverPage.jpg`) would have been an ideal thumbnail.

Fix: add a `--virtual-time-budget` to the command.

ricecooker/utils/images.py
rtibblesbot: Done in the latest push — added `--virtual-time-budget=8000` to the Chromium command (just before `--user-data-dir`). Chromium now advances its virtual clock to settle timers/network before the `--screenshot` capture, so JS-driven apps paint first instead of coming out blank white. And because the previous blank was a *successful* exit, this also restores the biggest-image fallback for the genuinely-nothing-rendered case.

rtibbles (commented):
Super - working as intended now! Can you please tidy up the commit history so we can merge?

rtibbles (approved):
Local testing checks out, all concerns addressed.

PR #721, opened 2026-07-24
Decompose IMS Content Packages (incl. SCORM) into a native node subtree (#332)

rtibblesbot:
## Summary

An IMS Content Package (including SCORM) could only be imported as one opaque HTML5 zip, so selecting a single node in Kolibri pulled the whole package, and plain media stayed trapped in a JS sandbox instead of playing as a native node. This migrates the `imscp` manifest parser into ricecooker and drives decomposition from the file-processing pipeline:

- `IMSCPConversionHandler` walks `imsmanifest.xml` into a tree-shaped `ContentNodeMetadata` and classifies each webcontent resource up a conservative ladder — native media node, else HTML5 zip or (when it qualifies) KPUB — rejecting assessment and QTI resources. Each surviving leaf re-enters the pipeline and is sealed into its own file rather than backed by the shared package zip.
- `ContentNode._process_uri` expands that tree into a Topic subtree, building each descendant through its own node constructor (`TreeNode.from_metadata`) and processing/validating the subtree leaves-first.
- `SCORM_metadata` (ported from ricecooker PR #468) maps the manifest's LOM metadata onto node fields — title, description, tags, language, learning activities, resource types, beginner level, and a license inferred from `<rights>`.
- `utils/scorm.py` holds the classifiers: SCORM LMS-API boilerplate detection (so a plumbing-only SCO does not read as interactive), assessment semantics, and single-media-wrapper detection.

## References

Implements learningequality/ricecooker#332. Parser migrated from learningequality/imscp. QTI item ingestion deferred to learningequality/ricecooker#337. Cross-node reference rewriting descoped per the issue's resolved design.

## Reviewer guidance

- `ricecooker/utils/pipeline/convert.py` — `IMSCPConversionHandler` is registered before `HTML5ConversionHandler` in the `FirstHandlerOnly` convert stage. Confirm an IMSCP `.zip` is claimed for decomposition rather than wrapped whole by the HTML5 handler.
- `ricecooker/utils/pipeline/convert.py` — `HTML5ConversionHandler._promote_to_kpub` applies to every HTML5 zip, not just IMSCP leaves, and it *strips* SCORM boilerplate scripts, `.css` members and stylesheet `<link>`s rather than letting them disqualify the zip (this flips an existing `test_transfer.py` assertion). Confirm both the preset change and the styling loss are acceptable for existing chefs that ship static HTML5 zips expecting `HTML5_ZIP`.
- `.pre-commit-config.yaml` — `check-added-large-files` is excluded for `tests/testcontent/imscp/`, which adds ~4.4 MB of real IMSCP/SCORM zip fixtures.
- `ricecooker/classes/nodes.py` — `_process_uri` adds and processes descendant nodes during the node's own `process_files()`, after `ChannelManager.process_tree` has snapshotted the node list. See the upload gap under Deviations.

## Deviations from the issue spec

- KPUB promotion is not scoped to IMSCP leaves — it runs for every HTML5 zip. Restricting it to the decomposition path would have meant a second, near-identical qualification branch.
- Promotion is lossy by design: SCORM boilerplate scripts and stylesheets are removed so a plumbing-only SCO still qualifies, so a promoted zip renders unstyled. Only genuine scripting or a non-boilerplate `.js` member keeps a zip as `HTML5_ZIP`.
- Decomposed leaf files are not uploaded yet. `ChannelManager.process_tree` snapshots the node list before calling `process_files()`, so descendants created during expansion never enter `file_map`; the subtree structure serializes but the leaf bytes are never diffed or uploaded. Verified locally: `process_tree()` returns `[]` for a tree-expanding node. The fix belongs in `managers/tree.py` and is not in this branch.

## AI usage

Used Claude Code to implement the migrated parser, classifiers, KPUB-promotion branch, and IMSCP handler from a pre-approved plan using test-driven development, following the existing pipeline handler patterns. Verified with the full test suite and `prek`/`ruff`.

---

</details>

<table><tr><td><a href="https://tibbs.io/status">@rtibblesbot</a></td><td><h3>🟡 Waiting for feedback</h3><em>Last updated: 2026-07-26 21:11 UTC</em></td></tr></table>


rtibbles (changes requested):
Oh, I assumed you would look _more_ at https://github.com/learningequality/ricecooker/pull/468 than at the archived imscp repository. Please make sure to go back to the prior work on ricecooker here to better shape the code conventions, implementation, and above all, the fidelity to the IMSCP spec.

rtibbles:
I do think including the metadata mapping in this pull request would be very helpful, actually.

ricecooker/classes/nodes.py
```diff
@@ -925,6 +930,103 @@ def _validate(self):
             self._validate_uri()
         super(ContentNode, self)._validate()
 
+    # File.__init__ has no **kwargs; any non-constructor key (path, license, ...)
```
rtibbles: Why not just add `**kwargs` to the file __init__?

ricecooker/classes/nodes.py
```diff
@@ -925,6 +930,103 @@ def _validate(self):
...
+
+    def _file_from_metadata(self, metadata_dict):
+        """Build a File from a file-metadata dict, tolerating extra keys."""
+        kwargs = {k: v for k, v in metadata_dict.items() if k in self._FILE_INIT_KEYS}
+        # Inherit the node's language unless the pipeline inferred one of its own
+        # (e.g. a subtitle language extracted from the file itself).
+        kwargs.setdefault("language", self.language)
+        return File(**kwargs)
+
+    # License-related fields live on the node's License object, not as plain
```
rtibbles: This special handling feels incorrect. Isn't this duplicating some of what already exists in __init__ methods? Can't we extract that into a method and reuse that?

ricecooker/classes/nodes.py
```diff
@@ -934,26 +1036,83 @@ def _process_uri(self):
...
+        if (
+            self.kind is not None
+            and content_metadata.get("kind", self.kind) != self.kind
+        ):
+            raise InvalidNodeException(
+                "Inferred kind is different from content node class kind."
+            )
+        self._apply_content_metadata(self, content_metadata)
+
+    def _expand_content_tree(self, content_metadata):
```
rtibbles: It feels like this logic belongs on TreeNode, rather than ContentNode? The differential behaviour/polymorphism _feels_ like it could be better expressed through inheritance rather than things being hardcoded into ContentNode.

ricecooker/utils/pipeline/convert.py
```diff
@@ -302,44 +324,121 @@ def _validate_index_html_body(self, zf, path, index_path="index.html"):
...
+    """Why an archive's members bar it from being a KPUB, or None when they do not."""
+    if any(n.lower().endswith(".js") for n in names):
+        return "JavaScript files (.js) are not allowed."
+    if any(n.lower().endswith(".css") for n in names):
+        return "external CSS files (.css) are not allowed."
+    return None
+
+
+def _kpub_disqualifier(names, index_html, entry="index.html"):
+    """The first reason a KPUB candidate fails the criteria; None ⇒ it qualifies.
```
rtibbles, line 346: Let's apply /terse-writing to all code comments introduced in this PR please. Concise, precise comments are the only ones allowed.

ricecooker/utils/pipeline/convert.py
```diff
@@ -302,44 +324,121 @@ def _validate_index_html_body(self, zf, path, index_path="index.html"):
...
+    def _qualifies_as_kpub(self, temp_dir):
+        """A static-article HTML5 zip is promoted to a KPUB; anything with scripts
+        or JS/CSS members stays an HTML5 zip.
+
+        Judged over the *processed* directory: reference resolution downloads
+        externally-referenced stylesheets and scripts into it, and those count
+        against the criteria just as much as ones the archive shipped with.
+        """
+        names = _archive_member_names(temp_dir)
+        # A .js/.css member disqualifies whatever the markup says, so test that
```
rtibbles: This seems to entirely contradict the stripping of SCORM specific stuff later in the file? Even then, just the existence of CSS is insufficient to rule out KPUBs - using a full HTML5 zip simply because of a little bit of unnecessary CSS styling is not helpful.

ricecooker/utils/pipeline/convert.py
```diff
@@ -302,44 +324,121 @@ def _validate_index_html_body(self, zf, path, index_path="index.html"):
...
-                audio_settings=audio_settings,
-                video_settings=video_settings,
-            )
+            super().handle_file(prepared_path, audio_settings, video_settings)
         finally:
             if prepared_path != path and os.path.exists(prepared_path):
                 os.unlink(prepared_path)
-        # Mirror Studio: when the entry point is not index.html at the root,
-        # record it in extra_fields.options.entry so Kolibri loads it.
+        # Mirror Studio: when the entry point is not index.html at the root, record
```
rtibbles: Why change this comment?

ricecooker/utils/pipeline/convert.py
```diff
@@ -525,26 +624,16 @@ def pre_process(self, temp_dir):
 
     def validate_archive(self, path: str):
         with self.open_and_verify_archive(path) as zf:
-            dom = self._validate_index_html_body(zf, path)
-
-            # Check for inline <script> tags (parsed without namespaces)
-            for _ in dom.iter("script"):
+            try:
+                index_html = zf.read("index.html")
```
rtibbles: Sorry, why have you changed the contract here? KPUB is allowed to have an entry specified: https://github.com/learningequality/kolibri/blob/develop/kolibri/plugins/safe_html5_viewer/frontend/views/SafeHtml5RendererIndex.vue#L72

But you've decided that we shouldn't allow this any more even though it's completely supported, why?

ricecooker/utils/pipeline/convert.py
```diff
@@ -751,6 +840,310 @@ def handle_file(self, path):
...
+        metadata = fm.content_node_metadata or {}
+        return metadata.get("kind"), files, metadata.get("extra_fields")
+    return None, files, None
+
+
+# The IMS Content Package manifest, always at the root of the package.
+_IMSCP_MANIFEST = "imsmanifest.xml"
+
+
+class _Package:
```
rtibbles: Feels like this could live in the imscp.py utils package and be imported?

ricecooker/utils/pipeline/convert.py
```diff
@@ -751,6 +840,310 @@ def handle_file(self, path):
...
+        }
+        if extra_fields:
+            leaf["extra_fields"] = extra_fields
+        _merge_lom_fields(leaf, fields)
+        return leaf
+
+    def _process_leaf(self, node_dict, package, index_html):
+        """Run the resource up the ladder and return its sub-pipeline result.
+
+        A resource that reduces to a single wrapped media file is processed as
```
rtibbles: Another very verbose comment.

ricecooker/utils/scorm.py
```diff
@@ -0,0 +1,201 @@
+"""Conservative classifiers for SCORM/IMSCP webcontent resources.
```
rtibbles, line 1: This doesn't belong in the pipeline module, it should be in the general utils folder.

ricecooker/utils/scorm.py
```diff
@@ -0,0 +1,201 @@
...
+
+# Recording a grade is assessment meaning whatever else the script does.
+SCORM_SCORE_RE = re.compile(r"cmi\.core\.score|cmi\.score", re.IGNORECASE)
+
+# Reporting a lesson status is only a signal outside SCORM plumbing: a plain
+# content SCO reports completion on unload as part of that plumbing.
+SCORM_STATUS_RE = re.compile(r"cmi\.core\.lesson_status", re.IGNORECASE)
+
+# HotPotatoes quiz engine globals. Their presence means the page IS an exercise.
+_HOTPOTATOES_GLOBALS_RE = re.compile(r"JQuiz|JCloze|JMatch|JMix|JCross", re.IGNORECASE)
```
rtibbles, line 46: What is the provenance of hot potatoes metadata?

rtibbles (changes requested):
Some concerns with architecture and overall structure, as well as a big concern about the regression for KPUB behaviour that you decided to add.

I also would like you to review the test suites you have written - please make sure they are focused on integration testing of input SCORM files to outputted ricecooker classes.

Any duplication, overly granular testing, or other redundant tests should be cleaned up.

ricecooker/classes/nodes.py
rtibblesbot: Done — `File.__init__` takes `**kwargs` now (`classes/files.py:73`), so a pipeline file-metadata dict splats straight in and `nodes.py` no longer keeps a copy of the parameter list. `_file_from_metadata` is down to inheriting the node language.

ricecooker/classes/nodes.py
rtibblesbot: Agreed — removed. Descendants are now built by their own constructors (`TreeNode.from_metadata`), which is where that metadata handling already lived, so there is no second setattr implementation.

For the one node that already exists when metadata arrives (the node whose `uri` was decomposed), there is `Node.set_metadata` — a single method applying a partial metadata dict, sharing `_set_license_fields` with `__init__` rather than reimplementing the license routing.

The "inferred license needs a copyright holder" special case moved out of `nodes.py` entirely to `utils/SCORM_metadata.py:_drop_unattributable_license`, which is where the inference happens: a license requiring attribution is only inferred when the LOM also names someone to attribute it to.

ricecooker/classes/nodes.py
rtibblesbot: Moved to `TreeNode` and made polymorphic:

- `TreeNode.from_metadata` picks the class (`TopicNode` when the metadata has no concrete `kind`, else `ContentNode`) and hands the non-structural keys to its constructor.
- Each class supplies its own `add_metadata_content`: `TreeNode` attaches children, `ContentNode` attaches files. No dict-shape branching in `ContentNode` any more.
- `expand_metadata_tree` / `process_metadata_subtree` are `TreeNode` methods, so any tree node can be the root of a decomposed subtree.

`ContentNode` keeps only what is genuinely `ContentNode`-specific: `INHERITED_METADATA_KEYS` adding `license`, and the `_validate` early return for a node that has become a folder.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Done in 9a23c9f — went through every comment and docstring this PR adds and cut the restatements and the padding. The multi-line ones that survive each carry one non-obvious fact.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Reverted — restored verbatim. It was reflowed when a sentence about promoted KPUBs was appended, and that sentence is gone now (a promoted KPUB can have an entry point, see the thread on `KPUBConversionHandler`).

ricecooker/utils/pipeline/convert.py
rtibblesbot: You are right on both counts, and they have the same fix: promotion now *removes* the things that are not content instead of refusing to promote because of them.

`HTML5ConversionHandler._promote_to_kpub` (`convert.py:485`):

- SCORM boilerplate scripts are stripped from the entry document **and** the wrapper `.js` members are deleted, so the discount and the shipped output finally agree. A static SCO wrapped in `SCORM_API_wrapper.js` is now a KPUB.
- `.css` members and their `<link rel=stylesheet>` are stripped too (new `references.strip_stylesheet_links`), so a bit of unnecessary styling no longer costs a full HTML5 zip.
- Only a non-boilerplate `.js` member or a surviving inline `<script>` keeps a zip as HTML5.

Still judged after reference resolution, so an externally-referenced stylesheet that was downloaded into the archive is stripped rather than sealed into the `.kpub`.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Not intended — thank you, that was a straight regression in the refactor. Both handlers now honour a specified entry.

`HTML5ConversionHandler` and `KPUBConversionHandler` share a new `WebArchiveConversionHandler`, which denests the zip, finds the entry with `find_html_entrypoint`, validates *that* document, and records it in `extra_fields.options.entry` when it is not a root `index.html`. So:

- a hand-authored `.kpub` whose entry is `content.html` validates and ships with the hint;
- an HTML5 zip whose static article is at `article.html` is promoted to KPUB and keeps its entry hint.

`sanitize_kpub_directory` takes the entry rather than assuming `index.html`. Tests: `TestKPUBValidation.test_non_index_entry_point_accepted` and `TestKPUBPromotion.test_non_index_entry_promoted_with_entry_hint`.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Moved — it is `IMSCPPackage` in `ricecooker/utils/imscp.py` now and imported by `convert.py`. The path-containment helper it depends on was already there.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Trimmed, along with the rest of the comments this PR adds (9a23c9f).

ricecooker/utils/scorm.py
rtibblesbot: Moved to `ricecooker/utils/scorm.py`.

ricecooker/utils/scorm.py
rtibblesbot: It comes from the migrated fixture. `tests/testcontent/imscp/test_quiz.zip` — one of the three examples in the archived `learningequality/imscp` repo, and the only assessment sample we have — is a Hot Potatoes export:

- `imsmanifest.xml` carries `<!-- JQuiz Version 7.0 Release 0 Build 33.-->`
- `test_quiz.htm` has `<meta name="author" content="Created with Hot Potatoes by Half-Baked Software, registered to kevino."/>` plus a `Hot Potatoes, Half-Baked Software, ...` keywords meta, and defines the `JQuiz` global.

So the signal is "this is the shape the one real quiz package we have takes", not something from the IMSCP/SCORM spec. Note it is matched on `<meta ... content="...Hot Potatoes...">` rather than `name="generator"`, because that is what the real file uses; `JQuiz`/`JCloze`/`JMatch`/`JMix`/`JCross` are the engine globals for its five exercise types.

Worth saying that it is belt-and-braces here: `test_quiz.htm` writes `cmi.core.score` from an inline script three times, so `has_assessment_semantics` rejects it on the score signal even with the Hot Potatoes checks removed (it declares no `adlcp:masteryscore`). Happy to drop the tool-specific checks if you would rather not special-case one authoring tool — the fixture still gets rejected, we would just lose the ability to catch a Hot Potatoes page that reports nothing back to the LMS.

rtibbles (changes requested):
It's been a long time for both of us since we looked at this. The PR is much closer to the _shape_ that I expected, but I have not gone through and done a detailed read through again. Before I do, please do a thorough self review of this work and update it as necessary.

rtibbles:
> Single-leaf organizations aren't collapsed.

I think it would be preferable to collapse to the minimal needed hierarchy (so a two deeply nested resources, would just be two resources in a single folder, and a single leaf would just be a resource by itself).

ricecooker/utils/pipeline/convert.py
```diff
@@ -386,8 +483,59 @@ def _prepare_archive(self, path):
...
+            return tmp_path, entry[len(prefix) :] if entry.startswith(prefix) else entry
+        return tmp_path, self.entry_point([n[len(prefix) :] for n in names])
+
+
+class HTML5ConversionHandler(WebArchiveConversionHandler):
+    EXTENSIONS = {file_formats.HTML5}
+    FILE_TYPE = "HTML5"
+
+    def seal_ext(self, temp_dir, ext, entry=None):
+        if self._promote_to_kpub(temp_dir, entry):
```
rtibbles: **blocking:** promotion ignores the context the typed callers already send (`decompose_packages=False` from `DownloadFile` and typed `_process_uri`). Gate `_promote_to_kpub` on a context flag and add a test in `tests/pipeline/test_convert.py` that a script-free zip stays `HTML5_ZIP` with it off.

`decompose_packages` doesn't describe promotion. Suggest one flag meaning "caller fixed the kind" (e.g. `preserve_kind`) replacing it, since both legacy paths set both.

ricecooker/utils/pipeline/convert.py
```diff
@@ -751,6 +886,216 @@ def handle_file(self, path):
...
+                    temp_dir, posixpath.dirname(_manifest_member(zf.namelist()))
+                )
+                zf.extractall(temp_dir)
+            try:
+                manifest = parse_imscp_manifest(ims_dir)
+            except ET.ParseError as e:
+                raise InvalidFileException(
+                    f"File {path} is not a valid IMSCP package, its {IMSCP_MANIFEST} could not be parsed: {e}"
+                )
+            children = self._build_nodes(
```
rtibbles: **suggestion:** when every resource is rejected (e.g. `tests/testcontent/imscp/test_quiz.zip`), the node becomes an empty topic with no log output, which would upload to Studio as-is. Warn, or fail the node.

rtibbles (changes requested):
Tested locally: `gitta_ims.zip` and `eventos.zip` decompose correctly through `ChannelManager.process_tree()` + `validate()`, and all 35 descendant files reach `file_map`.

Blocking: KPUB promotion breaks the legacy HTML5 APIs for any script-free zip (all pass on `main`):

| Caller | Result |
|---|---|
| `HTML5AppNode(files=[HTMLZipFile(zip)])` | `No required format preset found out of ('html5_zip',)` |
| `HTML5AppNode(uri=zip)` | `Inferred kind is different from content node class kind.` |

`HTML5AppNode` / `HTMLZipFile` state the author's intent and must stay `HTML5_ZIP`. `ContentNode(uri=zip)` promoting to KPUB is intended.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Done: `decompose_packages` replaced by one `preserve_kind` context flag, gating both KPUB promotion and IMSCP decomposition.

- Searched all 7 `.execute(` pipeline call sites and every convert handler that reclassifies output (`seal_ext`): only `HTML5ConversionHandler` does, and 2 callers fix a kind — `DownloadFile` and typed `ContentNode._process_uri`. Both now send `preserve_kind=True`.
- Tests in `TestKPUBPromotion`: script-free zip stays `HTML5_ZIP` with the flag; both rows of your table (`HTMLZipFile` and `HTML5AppNode(uri=…)`) stay `HTML5_ZIP`.
- Cache: preserved-kind keys match `main`; the promoting path gets a `:promote` suffix so pre-PR cached HTML5 output is not reused.
- On the CSS question (deleted thread): yes, intended — KPUB cannot carry stylesheets, and a chef that needs them now has `HTML5AppNode`/`HTMLZipFile` as a clean opt-out.

ricecooker/utils/pipeline/convert.py
rtibblesbot: Now fails the node: when no resource survives, the handler raises `InvalidFileException`, so `test_quiz.zip` logs an error and the node fails validation instead of uploading.

Same class one level down: a sub-folder whose resources were all rejected was kept as an empty topic (`_build_node`). Searched the other tree-building paths (`_build_nodes`, `collapse_single_children`); only `_build_node` emitted empty topics. It now drops the folder with a warning. Updated the 3 tests that asserted an empty tree, and extended `test_nested_resources_collapse_into_one_folder` with an all-rejected folder.

ricecooker/classes/nodes.py
```diff
@@ -700,6 +739,97 @@ class TreeNode(Node):
...
+            TopicNode
+            if metadata.get("kind", content_kinds.TOPIC) == content_kinds.TOPIC
+            else ContentNode
+        )
+        kwargs = {
+            key: value
+            for key, value in inherited.items()
+            if key in node_class.INHERITED_METADATA_KEYS
+        }
+        kwargs.update(cls.own_metadata_fields(metadata))
```
rtibbles: **blocking:** descendants take the raw manifest `identifier` as `source_id`, and `content_id` is `uuid5(channel domain, source_id)` (`nodes.py:840`). So unrelated packages that share identifiers collide. Two synthetic packages with items `ITEMA`/`ITEMB`, declared as `math-course` and `history-course`, gave identical content_ids for different content. Generated ids like gitta's `ITEM-d8e1` (DocBook `generate-id`) make this likely in real channels, and Kolibri keys progress on content_id.

Namespace descendant source_ids under the declaring node's source_id during expansion. Add a test that two packages with identical identifiers produce distinct content_ids.

ricecooker/classes/nodes.py
```diff
@@ -700,6 +739,97 @@ class TreeNode(Node):
...
+        fields = self.own_metadata_fields(metadata)
+        is_lone_leaf = (
+            len(children) == 1
+            and children[0].get("kind", content_kinds.TOPIC) != content_kinds.TOPIC
+        )
+        if is_lone_leaf:
+            fields.update(self.own_metadata_fields(children[0]))
+        # The node keeps its own identity.
+        fields.pop("source_id", None)
+        self.set_metadata(fields)
```
rtibbles: **suggestion:** this replaces the chef's `title` with the manifest's. `ContentNode(title="math-course", uri=<two-item package>)` ends up titled `"Org"` (the organization title), and gitta ends up as `"Introduction DBS"`. The LOM license overriding the chef's is tested and intended. Is the title meant to as well? If not, let the chef's explicit fields win and use the manifest only as a fallback.

ricecooker/utils/imscp.py
```diff
@@ -0,0 +1,498 @@
...
+
+
+def _element_text(elem):
+    """Concatenate all descendant text/tail (ignoring ``<br>``), stripped."""
+    if elem is None:
+        return ""
+    return "".join(elem.itertext()).strip()
+
+
+def contained_path(root, member):
```
rtibbles: **praise:** every untrusted manifest path (href, `<file>`, metadata `location`, wrapped media) goes through this one containment check, and the traversal test covers it.

rtibbles (changes requested):
### Prior findings

All prior findings resolved.

**Newly resolved:**
- Legacy HTML5 APIs broken by KPUB promotion (blocking): `preserve_kind` from `DownloadFile` and typed `_process_uri`. `HTML5AppNode(uri=)` and `HTML5AppNode(files=[HTMLZipFile])` both stay `html5_zip`, checked locally
- Empty topic for all-rejected package (suggestion): raises `InvalidFileException`, and emptied sub-folders are dropped. `test_quiz.zip` fails the node

### This round

CI passing. Checked locally through `ChannelManager.process_tree()`: gitta (28 leaves) and eventos (7) validate, serialize, and all leaf files reach the upload list. `ContentNode(uri=static.zip)` → KPUB.

- **blocking:** descendant content_ids collide across packages (inline)
- **suggestion:** manifest title overwrites the chef's title (inline)

ricecooker/classes/nodes.py
rtibblesbot: Fixed in 04b14f7d: `from_metadata` prefixes every descendant `source_id` with the declaring node's (`math-course/ITEMA`). A lone-leaf package keeps the chef's `source_id` unchanged.
- Searched for other places decomposer ids become `source_id`: `from_metadata` is the only constructor, and `expand_metadata_tree` is its only entry point, so 1 match, changed.
- Test: `test_packages_sharing_identifiers_expand_distinctly` builds your two-package case and asserts 4 distinct content_ids.

ricecooker/classes/nodes.py
rtibblesbot: Not intended. Fixed in 04b14f7d: in `expand_metadata_tree`, a manifest field applies only when the node's own value is empty. License fields still override, and `extra_fields` still merges.
- Searched `set_metadata` callers: 2. This one is changed. The other, the non-decomposed `_process_uri` path, already overwrote chef fields with pipeline metadata on `main`, so I left it alone.
- The same test asserts that both nodes keep the chef titles.

ricecooker/utils/imscp.py
rtibblesbot: Thanks.

rtibbles (approved):

PR #739, opened 2026-09-17
Fix add_nodes dropping sibling subtrees when a chunk's last child is skipped

amritraj7237:
### What's wrong

`add_nodes()` sends children to Studio in chunks of 10, skipping any child whose primary file failed to upload or that didn't validate. The recursion into those children was guarded by:

```python
if response_json["root_ids"].get(child.get_node_id().hex):
    for child in chunk:
        self.add_nodes(...)
```

By the time that `if` runs the loop above has finished, so `child` is still the last item in the chunk. The guard asks whether that one child was created and then applies the answer to all ten. When the last child is a skipped one its node_id was never sent, so it's absent from `root_ids`, the guard is false, and none of the chunk's subtrees are uploaded.

It reports success while doing it. Only the skipped node goes into `failed_node_builds`, so `check_failed()` still prints "All nodes were created successfully", and `upload_count += len(chunk)` counts the skipped child, so progress still reaches 100%.

I ran into it through `failed_uploads`, which is easy to hit on a long run. With a topic holding 9 subtopics plus one leaf whose upload failed, the 9 subtopics were created and all of their contents were missing.

### Changes

* Recurse per child, over only the children actually put in the payload. This also stops `add_nodes(None, child)` being called for skipped children, which posted their children with `root_id: None`.
* `upload_count` counts what was sent instead of the whole chunk.
* `except ConnectionError` never caught anything. `ConnectionError` isn't imported in this module so it resolved to the builtin, and requests' version subclasses `RequestException`/`OSError`, not the builtin. Switched to `RequestException`.
* With that handler actually working, a dropped connection would be recorded and the channel committed anyway with subtrees missing. Whole-batch failures now go in `failed_batches` and `upload_tree()` raises `ChannelIncompleteError` rather than committing. Individual node failures still commit as before.

That last point also changes the existing non-200 branch, which committed partial trees when Studio returned a 500. I can put it behind a flag if you'd rather not change that in this PR.

### Tests

Four added, one corrected. `test_add_nodes_handles_connection_error` was raising the builtin `ConnectionError`, which requests never raises, so it passed against the broken handler. `test_add_nodes_skips_invalid_nodes` already had the triggering shape (invalid child last) but only asserted on the payload, never that the valid sibling was recursed into, which is why this went unnoticed.

I couldn't get a complete local environment running (no ffmpeg, pandoc or poppler), so please confirm against CI.


ricecooker/managers/tree.py
```diff
@@ -416,19 +444,25 @@ def add_nodes(self, root_id, current_node, indent=1):  # noqa: C901
...
-                                child,
-                                indent + 1,
-                            )
-        except ConnectionError as ce:
-            self.failed_node_builds[root_id] = {"node": current_node, "error": ce}
+                    self.node_count_dict["upload_count"] += len(sent_children)
+
+                    root_ids = response_json["root_ids"]
+                    for child in sent_children:
+                        child_root_id = root_ids.get(child.get_node_id().hex)
```
rtibblesbot, line 489: ✅ **Resolved** — addressed in the current code.


**suggestion:** A child that was sent but is absent from `root_ids` is still silently dropped.

The POST returned 200 and Studio handed back no id for it. The guard skips it, so its whole subtree is never uploaded. Nothing lands in `failed_node_builds` or `failed_batches`. `check_failed()` then prints "All nodes were created successfully" and the channel commits with the subtree missing.

The guard itself is load-bearing — you can't recurse with `None`. Add the missing `else` that records the child, so the commit refusal covers it.

Is a partial `root_ids` actually reachable, or is the `.get()` purely defensive?

ricecooker/managers/tree.py
```diff
@@ -405,6 +432,7 @@ def add_nodes(self, root_id, current_node, indent=1):  # noqa: C901
                             }
                     else:
                         payload_children.append(child.to_dict())
+                        sent_children.append(child)
                 payload = {"root_id": root_id, "content_data": payload_children}
```
rtibblesbot, line 471: ✅ **Resolved** — addressed in the current code.


**suggestion:** A chunk whose children are all skipped can now refuse the whole channel. `payload_children` is `[]`, and this still POSTs `content_data: []`. A non-200 on that request appends to `failed_batches`, and `upload_tree` raises `ChannelIncompleteError`. The batch carried nothing, so that is exactly the "individual node failed, still commit" case this PR says it preserves.

Add `if not payload_children: continue` before the POST. That drops both the round trip and the escalation risk.

I haven't read Studio's `add_nodes` view — does it 200 an empty `content_data`?

ricecooker/managers/tree.py
```diff
@@ -301,6 +306,14 @@ def upload_tree(self):
 
         self.add_nodes(root, self.channel)
         self.check_failed()
+        if self.failed_batches:
+            # Committing now would stage a channel with whole subtrees missing
+            # and nothing but a warning to say so.
+            raise ChannelIncompleteError(
```
rtibblesbot, line 314: ✅ **Resolved** — addressed in the current code.


**suggestion:** `ChannelIncompleteError` reaches the chef operator as an uncaught traceback. Nothing between here and `main` catches it. The operator gets a stack trace on top of the report `check_failed()` has already printed.

`commands.py:109-112` is the repo's pattern for a known, already-logged failure: log, then `sys.exit(1)`. Catch this in `uploadchannel` and exit the same way.

ricecooker/managers/tree.py
```diff
@@ -416,19 +444,25 @@ def add_nodes(self, root_id, current_node, indent=1):  # noqa: C901
                         "error": response.reason,
                         "content": response.content,
                     }
+                    self.failed_batches.append(root_id)
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**suggestion:** A `root_id` that fails in two chunks loses all but the last error. Both failure paths assign `failed_node_builds[root_id]`, so the earlier entry is clobbered. Both also append the same `root_id`, so `failed_batches` holds duplicates and `check_failed()` prints the identical line twice. Across paths the loss is worse: a non-200 stores the response body, and a later `RequestException` replaces that entry with one that has none.

Store the error on the `failed_batches` entry itself. That makes the report self-contained and drops the lookup-by-key indirection.

Separately, the loop keeps going after a failed batch. Every remaining chunk is posted and every other branch recursed into, and `upload_tree` then discards all of it. Collecting every failure for one report has value — tell me which behaviour you intend, and leave a comment there if it's deliberate.

ricecooker/managers/tree.py
```diff
@@ -314,6 +327,16 @@ def truncate_fields(self, node):
             self.truncate_fields(child)
 
     def check_failed(self):
+        if self.failed_batches:
+            config.LOGGER.error(
+                "ERROR: {} node batch(es) failed to upload. Every node under "
+                "them is missing from the channel:".format(len(self.failed_batches))
+            )
+            for root_id in self.failed_batches:
```
rtibblesbot: ✅ **Resolved** — addressed in the current code.


**nitpick:** A failed batch is reported three times. This block logs it. The `failed_node_builds` loop below re-logs the same entry as a warning. `upload_tree()` then raises with a third copy of "N node batch(es) failed to upload". Drop one of the first two.

ricecooker/exceptions.py
```diff
@@ -71,6 +71,20 @@ def __init__(self, *args, **kwargs):
...
 
+class ChannelIncompleteError(Exception):
+    """ChannelIncompleteError: raised when whole batches of nodes failed to
+    upload, so committing would stage a channel with subtrees missing.
+
+    Distinct from an individual node failing to build (a bad file, a node that
+    did not validate): those are reported and the channel is still committed.
+    This is raised only when an entire add_nodes request never landed, which
+    takes every descendant of that request with it.
+    """
```
rtibblesbot, line 82: ✅ **Resolved** — addressed in the current code.


**nitpick:** This `__init__` adds nothing over inheriting. `Exception.__init__` raises `TypeError` on any keyword argument, so the `**kwargs` is misleading. Drop it — it exists only to match `InvalidQuestionException` above, and that pattern isn't worth propagating.

tests/test_tree.py
```diff
@@ -1161,17 +1222,151 @@ def test_add_nodes_handles_connection_error(channel):
...
+
+    with patch("ricecooker.config.SESSION.post", side_effect=fake_post):
+        manager.upload_tree()
+
+    assert manager.failed_batches == []
+    assert bad.get_node_id().hex in manager.failed_node_builds
+    assert config.finish_channel_url() in committed
+
+
+def test_add_nodes_records_a_failed_batch_on_server_error(channel):
```
rtibblesbot: **nitpick:** This test duplicates `test_add_nodes_handles_server_error`. Same mocks, same 500 `MagicMock`, differing only by the `failed_batches == ["root_id"]` assertion. Move that assertion into the existing test and delete this one — same coverage, 24 fewer lines.

Same shape at line 1348. `test_add_nodes_handles_read_timeout` enters the identical `except RequestException` block as `test_add_nodes_handles_connection_error`. If the point is that the catch is now the `requests` base class, parametrize one test body over both exception classes instead.

ricecooker/managers/tree.py
```diff
@@ -416,19 +444,25 @@ def add_nodes(self, root_id, current_node, indent=1):  # noqa: C901
...
-        except ConnectionError as ce:
-            self.failed_node_builds[root_id] = {"node": current_node, "error": ce}
+                    self.node_count_dict["upload_count"] += len(sent_children)
+
+                    root_ids = response_json["root_ids"]
+                    for child in sent_children:
+                        child_root_id = root_ids.get(child.get_node_id().hex)
+                        if child_root_id:
+                            self.add_nodes(child_root_id, child, indent + 1)
+        except RequestException as e:
```
rtibblesbot, line 503: ✅ **Resolved** — addressed in the current code.


**praise:** Correct diagnosis. `requests.exceptions.ConnectionError` isn't a subclass of the builtin, so the old handler was dead code. The old test passed only because it raised the builtin. The comment explains why the obvious-looking name was wrong, which is where a comment earns its place.

rtibblesbot (commented):
PR #739 — the fix is correct and the regression tests pin it. Reverting `tree.py` to `upstream/main` fails all five new/corrected tests. `tests/test_tree.py`: 64 passed. `uvx prek run --all-files` clean. CI pending at review time.

Suggestions (inline):
- A child missing from `root_ids` is still silently dropped — the same failure class this PR fixes.
- An all-skipped chunk still POSTs an empty `content_data`, and a non-200 there now refuses the whole channel.
- `ChannelIncompleteError` reaches the chef operator as an uncaught traceback.
- A `root_id` that fails in two chunks loses all but the last error.

Nitpicks on the triple-reported batch failure, the no-op `__init__`, and two near-duplicate tests.

---

Ran a phased review pipeline over the pull request diff:
- Classified the diff to select review passes (core, frontend, backend) and whether manual QA was required
- Core review pass checked correctness, design, architecture, testing, completeness, and DRY/SRP/Rule-of-Three principles
- Specialized frontend/backend review passes applied framework-specific lenses where those files changed
- For UI changes: manual QA and an accessibility audit against a live dev server, when available
- Checked CI status and linked issue acceptance criteria
- Synthesized one review from those passes and chose the verdict from the findings, CI status, and QA evidence

</details>

ricecooker/managers/tree.py
amritraj7237: Added the `else`. It records the child via `_record_failed_batch`, so a child Studio returns no id for now blocks the commit instead of vanishing.

On your question: it's defensive. I haven't read Studio's `add_nodes` view, so I can't tell you whether a partial `root_ids` is reachable in practice — the `.get()` was in the original code and I kept it. Either way, recording beats skipping: if it's unreachable the branch costs nothing, and if it isn't, it's the exact silent loss this PR exists to stop. Happy to drop it to an assertion if you know the endpoint always echoes every id.

ricecooker/managers/tree.py
amritraj7237: Good catch, this contradicted the PR description. Added `if not payload_children: continue` before the POST.

I haven't read the Studio view either, so I don't know whether it 200s an empty `content_data` — but the request has nothing to send regardless, so skipping it removes the round trip and the escalation risk without needing the answer.

ricecooker/managers/tree.py
amritraj7237: Done. `uploadchannel` now catches `ChannelIncompleteError`, logs it and calls `sys.exit(1)`, matching the `add_channel` handling at `commands.py:109`. `check_failed()` has already printed the detail by then.

ricecooker/managers/tree.py
amritraj7237: Moved both failure paths onto a `_record_failed_batch()` helper. `failed_batches` entries now carry `{root_id, node, error}`, so two failures under one parent both survive and the report needs no lookup. `failed_node_builds[root_id]` is still written — `test_add_nodes_handles_server_error` asserts on it — but nothing reads it for the report.

Also removed the triple reporting you flagged separately, which exposed a related bug: `check_failed()` could print "All nodes were created successfully" when only batches had failed.

On the loop: continuing is deliberate. One run should surface every failure, not just the first, so an operator fixes them in one pass rather than rediscovering them one re-run at a time. Now stated in a comment above the loop.

ricecooker/managers/tree.py
amritraj7237: Dropped the first one. `check_failed()` now reports batch failures from `failed_batches` and filters those root_ids out of the `failed_node_builds` loop, so each batch is reported once. The raise message no longer restates the list, just the count and the reason.

Fixing this turned up a related bug in the same block: the `else` on the node-builds loop meant "All nodes were created successfully" could still print when only batches had failed. That branch is now `elif not self.failed_batches`.

ricecooker/exceptions.py
amritraj7237: Removed. You're right that it only existed to match the surrounding classes, and the `**kwargs` was misleading given `Exception.__init__` rejects keyword arguments.

tests/test_tree.py
amritraj7237: Both done. Deleted `test_add_nodes_records_a_failed_batch_on_server_error` and moved its `failed_batches` assertion into the existing `test_add_nodes_handles_server_error`.

The two transport tests are now one `test_add_nodes_handles_transport_errors`, parametrized over `requests.exceptions.ConnectionError` and `ReadTimeout` with ids `connection_error` / `read_timeout`. That is the point, so the parametrization says it better than two bodies did.

Tests are net +82/-64 overall, since I added coverage for the two behaviours from your suggestions above.

rtibblesbot (commented):
PR #739 — all 8 prior findings are settled. One new suggestion is inline.

CI passing. No UI files changed, so no visual verification.

- **suggestion** — `ricecooker/managers/tree.py:351`: the de-duplicated batch report no longer logs Studio's response body. That body is the only detail distinguishing one 500 from another. Suggested fix is in the inline comment.

<details><summary>Prior-finding status</summary>

RESOLVED — ricecooker/managers/tree.py:485 — child missing from `root_ids` silently dropped
RESOLVED — ricecooker/managers/tree.py:467 — all-skipped chunk POSTs empty `content_data`
RESOLVED — ricecooker/managers/tree.py:314 — `ChannelIncompleteError` reaches operator as traceback
RESOLVED — ricecooker/managers/tree.py:343 — `root_id` failing in two chunks loses all but last error
ACKNOWLEDGED — ricecooker/managers/tree.py:422 — loop continues after a failed batch (deliberate; now commented)
RESOLVED — ricecooker/managers/tree.py:351 — failed batch reported three times
RESOLVED — ricecooker/exceptions.py:82 — no-op `__init__` on `ChannelIncompleteError`
RESOLVED — tests/test_tree.py — two near-duplicate test pairs

</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Retrieved prior bot reviews via the GitHub API
- Classified each prior finding as RESOLVED, UNADDRESSED, ACKNOWLEDGED, or CONTESTED
- Only raised NEW findings for newly introduced code
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested
- Synthesized one review from the passes and chose the verdict from the findings, CI status, and QA evidence

</details>

ricecooker/managers/tree.py
```diff
@@ -313,17 +328,47 @@ def truncate_fields(self, node):
...
+        self.failed_batches.append({"root_id": root_id, **entry})
+
     def check_failed(self):
-        if len(self.failed_node_builds) > 0:
+        if self.failed_batches:
+            config.LOGGER.error(
+                "ERROR: {} node batch(es) failed to upload. Every node under "
+                "them is missing from the channel:".format(len(self.failed_batches))
+            )
+            for batch in self.failed_batches:
```
rtibblesbot, line 351: ✅ **Resolved** — addressed in the current code.


**suggestion:** Studio's response body no longer reaches the operator when a batch fails. The batch loop here prints only the root_id and the HTTP reason phrase. The loop below, which used to print the body, now filters batch root_ids out. The body is still recorded on the failure entry. Two different 500s therefore render identically as `under <id>: Internal Server Error`.

Cutting the triple report was my request and was right. This is collateral from where the cut landed, not a reason to restore the second loop. Please print the body on the batch line:

```python
for batch in self.failed_batches:
    config.LOGGER.error("\tunder {}: {}".format(batch["root_id"], batch["error"]))
    if "content" in batch:
        config.LOGGER.error("\t\t{}".format(batch["content"][:80]))
```

ricecooker/managers/tree.py
```diff
@@ -313,17 +328,47 @@ def truncate_fields(self, node):
...
-                node = self.failed_node_builds[node_id]
+            for node_id in node_builds:
+                node = node_builds[node_id]
                 config.LOGGER.warning(
                     "\t{} ({})".format(str(node["node"]), node["error"])
                 )
                 if "content" in node:
                     config.LOGGER.warning(node["content"][:80])
-        else:
+        elif not self.failed_batches:
```
rtibblesbot, line 375: ✅ **Resolved** — addressed in the current code.


**praise:** `elif not self.failed_batches` fixes a bug that predates this PR. `check_failed()` used to report "All nodes were created successfully." whenever every failure was a batch failure. That is the case where the channel is most incomplete.

ricecooker/managers/tree.py
amritraj7237: Confirmed and fixed — applied your snippet.

I reproduced it first: with a 500 carrying a distinct body, `check_failed()` printed only `under root_id: Internal Server Error` and the body never surfaced, though it was sitting on the entry. Two different 500s were indistinguishable, exactly as you describe.

Added `test_check_failed_logs_the_response_body_for_a_batch_failure` to pin it — it asserts both the reason phrase and the body reach the log, and fails without the fix.

One thing I left alone: `content` is `bytes`, so the line renders as `b'...'`. That matches what the old second loop did, so this restores the previous output rather than changing it. Happy to decode it if you'd prefer the cleaner form.

tests/test_tree.py
```diff
@@ -1145,33 +1150,261 @@ def test_add_nodes_skips_invalid_nodes(channel):
...
+
+
+def test_check_failed_logs_the_response_body_for_a_batch_failure(channel, caplog):
+    """Studio's response body has to reach the operator.
+
+    The HTTP reason phrase is the same for every 500; the body is the only
+    thing that distinguishes one from another.
+    """
+    manager = ChannelManager(channel)
+    manager._record_failed_batch(
```
rtibblesbot, line 1313: ✅ **Resolved** — addressed in the current code.


**suggestion:** No test asserts that the response body actually reaches the failed-batch record through `add_nodes`. This test passes `content` in by hand, so deleting `content=response.content` at `tree.py:481` would leave it green. `test_add_nodes_handles_server_error` does drive the real path, and already sets a response body. Add one assertion at the end of that test:

```python
assert manager.failed_batches[0]["content"] == b"Server error"
```

rtibblesbot (commented):
PR #739: all 10 prior findings are resolved. One new suggestion on the added test is inline.

<details><summary>Prior-finding status</summary>

RESOLVED — ricecooker/managers/tree.py:489 — Child sent but absent from `root_ids` silently dropped
RESOLVED — ricecooker/managers/tree.py:471 — Chunk with all children skipped could refuse the whole channel
RESOLVED — ricecooker/managers/tree.py:314 — `ChannelIncompleteError` reached the operator as a traceback
RESOLVED — ricecooker/managers/tree.py — `root_id` failing in two chunks lost all but the last error
RESOLVED — ricecooker/managers/tree.py — Failed batch reported three times
RESOLVED — ricecooker/exceptions.py:82 — `__init__` added nothing over inheriting
RESOLVED — ricecooker/managers/tree.py:503 — `requests.exceptions.ConnectionError` not a builtin subclass
RESOLVED — tests/test_tree.py — Duplicate of `test_add_nodes_handles_server_error` (verified: one 500-mock test remains; the new `check_failed` test mocks no session)
RESOLVED — ricecooker/managers/tree.py:351 — Studio's response body now reaches the operator
RESOLVED — ricecooker/managers/tree.py:375 — `elif not self.failed_batches` fix
</details>


---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Retrieved prior bot reviews via the GitHub API
- Classified each prior finding as RESOLVED, UNADDRESSED, ACKNOWLEDGED, or CONTESTED
- Only raised NEW findings for newly introduced code
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested
- Synthesized one review from the passes and chose the verdict from the findings, CI status, and QA evidence

</details>

tests/test_tree.py
amritraj7237: Added the assertion — but the premise isn't quite right, so flagging it.

I mutation-tested both paths:

**Dropping `content=response.content` at tree.py:481** does not leave the suite green. `test_add_nodes_handles_server_error` already ends with `assert manager.failed_node_builds["root_id"]["content"] == b"Server error"`, which fails with `KeyError: 'content'` under exactly that mutation. You're right that my new test stays green on its own, but the suite catches it.

**Putting `content` on `failed_node_builds` but not the batch entry** is caught by the new test, since `check_failed()` reads the body off the batch entry.

So both mutations were already covered, just split across two tests. I've added your assertion anyway: `check_failed()` reads `batch["content"]`, and pinning that in the test that drives the real `add_nodes` path makes the coverage direct instead of split. One line, no downside.

rtibblesbot (approved):
PR #739: all 11 prior findings resolved. No new findings. CI passing.

<details><summary>Prior-finding status</summary>

RESOLVED — ricecooker/managers/tree.py:489 — child sent but absent from `root_ids` silently dropped
RESOLVED — ricecooker/managers/tree.py:471 — all-skipped chunk refuses the whole channel
RESOLVED — ricecooker/managers/tree.py:314 — `ChannelIncompleteError` reaches operator as uncaught traceback
RESOLVED — ricecooker/managers/tree.py — `root_id` failing in two chunks loses earlier errors
RESOLVED — ricecooker/managers/tree.py — failed batch reported three times
RESOLVED — ricecooker/exceptions.py:82 — redundant `__init__`
RESOLVED — ricecooker/managers/tree.py:503 — `requests.exceptions.ConnectionError` diagnosis (praise)
RESOLVED — ricecooker/managers/tree.py:351 — Studio response body no longer reaches operator
RESOLVED — ricecooker/managers/tree.py:375 — `elif not self.failed_batches` fix (praise)
RESOLVED — tests/test_tree.py — test duplicates `test_add_nodes_handles_server_error`
RESOLVED — tests/test_tree.py:1313 — no test asserts response body reaches failed-batch record via `add_nodes`

</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Retrieved prior bot reviews via the GitHub API
- Classified each prior finding as RESOLVED, UNADDRESSED, ACKNOWLEDGED, or CONTESTED
- Only raised NEW findings for newly introduced code
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested
- Synthesized one review from the passes and chose the verdict from the findings, CI status, and QA evidence

</details>

rtibbles (approved):
Thank you @amritraj7237 - this is a helpful quality of life improvement!

If you notice anything else in future, please do feel free to open an issue, and I'll be happy to assign you.

